PR #2, opened 2017-05-11
remove interfaces, classes in support of db inserts

nicpottier:


server.go
```diff
@@ -321,11 +318,13 @@ func (s *server) handleIndex(w http.ResponseWriter, r *http.Request) {
 
 // RegisterHandler adds a new handler for a channel type, this is called by individual handlers when they are initialized
 func RegisterHandler(handler ChannelHandler) {
-	registeredHandlers[handler.ChannelType()] = handler
+	if handlers == nil {
```
rowanseymour: Needs reverted

PR #3, opened 2017-05-12
Insert msgs

nicpottier:


channel.go
```diff
@@ -181,13 +181,13 @@ const localTTL = 60
...
-	Address     string      `json:"address"       db:"address"`
-	Country     string      `json:"country"       db:"country"`
-	Config      string      `json:"config"        db:"config"`
+	OrgID       OrgID          `json:"org_id"        db:"org_id"`
+	ID          ChannelID      `json:"id"            db:"id"`
+	UUID        ChannelUUID    `json:"uuid"          db:"uuid"`
+	ChannelType ChannelType    `json:"channel_type"  db:"channel_type"`
+	Address     sql.NullString `json:"address"       db:"address"`
+	Country     sql.NullString `json:"country"       db:"country"`
+	Config      sql.NullString `json:"config"        db:"config"`
```
rowanseymour: Not so related to this PR, but might be simpler if this was a map and the JSON parsing happened during the DB fetch. It's kind weird that the JSON serialized version of a channel will have serialized JSON as a field

courier.toml
```diff
@@ -15,7 +15,7 @@ base_url = "https://localhost:8080"
 port = 8080
 
 # Our database connection string, right now only Postgres is supported
-db = "postgres://courier@localhost/courier?sslmode=disable"
+db = "postgres://courier@localhost/courier_test?sslmode=disable"
```
rowanseymour: No biggie but it was intentional having this not be the db used for unit tests

testdata.sql
```diff
@@ -0,0 +1,28 @@
+
+/* Org with id 100 */
+DELETE FROM orgs_org where id = 100;
```
rowanseymour: Seems odd having explicit deletes in a test data file - shouldn't we just truncate these tables?

msg.go
```diff
@@ -43,6 +38,38 @@ const (
...
+
+// MsgVisibility is the visibility of a message
+type MsgVisibility string
+
+// Possible values for MsgVisibility
+const (
+	MsgVisible       MsgVisibility = "V"
+	MsgDeleted       MsgVisibility = "D"
+	MsgArchived      MsgVisibility = "A"
+	NilMsgVisibility MsgVisibility = ""
```
rowanseymour: Do we need a nil value - shouldn't new Msgs get the db default like for priority?

nicpottier: This is Nil at the code level not at the DB level.

testdata.sql
nicpottier: Seems like only affecting what matters to us is more friendly

courier.toml
nicpottier: ya, could go either way here really

PR #4, opened 2017-05-16
new backend architecture

nicpottier:


backends/rapidpro/msg.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	// finally try to add this message to our handling queue
+	err = addToHandleQueue(b, m)
+
+	// TODO: spool backdown for failure to add to redis
+	return err
+}
+
+func newMsgFromMsg(m *courier.Msg) *Msg {
```
rowanseymour: What if we called RapidPro's message something like `DbMsg` to avoid this sort of confusion?

courier.toml
```diff
@@ -1,5 +1,6 @@
 #-------------------------------------------------------------------------------------
-# Courier settings can be set in four different wants, in order of preference:
+# Courier settings can be set in four different wants, in order of preference, last
```
rowanseymour: *ways

backends/rapidpro/msg.go
rowanseymour: Different backends don't interact so I'd be less worried about two different `DbMsg` names, but your call.

Issue #15, opened 2017-07-10
Dedupe incoming messages


nicpottier:
If we receive a message on the same channel, for the same contact and at the same time, then we should dedupe that and not recreate it.

nicpottier:
It is probably fair to only do this within a small time window. (say five mins)

nicpottier:
fixed

PR #17, opened 2017-07-10
change Msg to be an interface implemented by Backends, remove passthr…

nicpottier:
…oughs on Server to go straight to Backend


nicpottier:
Fixes https://github.com/nyaruka/courier/issues/15

PR #30, opened 2017-08-03
Zenvia handlers

norkans7:


handlers/base.go
```diff
@@ -189,3 +190,12 @@ func DecodePossibleBase64(original string) string {
 
 	return decoded
 }
+
+// EncodeBase64 encodes the list of strings with Linux base64
+func EncodeBase64(strList []string) string {
```
rowanseymour: I might be wrong but isn't this method just concatenating strings? I know the json marshaler will use base64 for byte[] but I think we should do an explicit conversion.

norkans7: my bad on this

norkans7: Fixed it

handlers/zenvia/zenvia.go
```diff
@@ -1,3 +1,232 @@
...
+//         	"statusMessage": "Delivered",
+//         	"statusDetail": "120",
+//         	"statusDetailMessage": "Message received by mobile",
+//         	"id": "hs765939216",
+//         	"received": "2014-08-26T12:55:48.593-03:00",
+//         	"mobileOperatorName": "Claro"
+// 		}
+// }
+type statusRequest struct {
+	CallbackMtRequest struct {
```
nicpottier: Acronyms should always be all caps in Go. So should be `CallbackMTRequest` (you can use the `json` tag to remap the name)

handlers/base.go
nicpottier: This feels more like something that should live in `utils`.

handlers/zenvia/zenvia.go
norkans7: this is not clear to me

nicpottier: Which part? Just saying class names shouldn't have mixed acronym case in Go. So instead of `Mt` you should always use `MT`, just like for `http` should always be `HTTP`

norkans7:
This has been updated and can be merged after #54

handlers/zenvia/zenvia.go
```diff
@@ -1,3 +1,234 @@
...
+		return nil, fmt.Errorf("no username set for Zenvia channel")
+	}
+
+	password := msg.Channel().StringConfigForKey(courier.ConfigPassword, "")
+	if password == "" {
+		return nil, fmt.Errorf("no password set for Zenvia channel")
+	}
+
+	encodedCreds := base64.StdEncoding.EncodeToString([]byte(strings.Join([]string{username, ":", password}, "")))
+	authHeader := "Basic " + encodedCreds
```
nicpottier: Just realized that for these (and for InfoBip) should just be using:

```
req.SetBasicAuth(username, password)
```

handlers/zenvia/zenvia.go
```diff
@@ -1,3 +1,231 @@
...
+	"07": courier.MsgErrored,
+	"08": courier.MsgErrored,
+	"09": courier.MsgErrored,
+	"10": courier.MsgErrored,
+}
+
+// Initialize is called by the engine once everything is loaded
+func (h *handler) Initialize(s courier.Server) error {
+	h.SetServer(s)
+	err := s.AddHandlerRoute(h, "POST", "receive", h.ReceiveMessage)
```
nicpottier: use `http.MethodPost`. Also move this to the top after `NewHandler`

handlers/zenvia/zenvia.go
```diff
@@ -1,3 +1,231 @@
...
+	return s.AddHandlerRoute(h, "POST", "status", h.StatusMessage)
+}
+
+// ReceiveMessage is our HTTP handler function for incoming messages
+func (h *handler) ReceiveMessage(ctx context.Context, channel courier.Channel, w http.ResponseWriter, r *http.Request) ([]courier.Event, error) {
+	// get our params
+	zvMsg := &messageRequest{}
+	err := handlers.DecodeAndValidateJSON(zvMsg, r)
+	if err != nil {
+		return nil, err
```
nicpottier: should write using `WriteAndLogRequestError` not return the err

PR #33, opened 2017-08-16
Sq handlers

norkans7:


handlers/shaqodoon/shaqodoon.go
```diff
@@ -1,6 +1,210 @@
...
+	s.AddUpdateStatusRoute(h, "GET", "sent", sentHandler)
+	s.AddUpdateStatusRoute(h, "POST", "sent", sentHandler)
+
+	deliveredHandler := h.buildStatusHandler("delivered")
+	s.AddUpdateStatusRoute(h, "GET", "delivered", deliveredHandler)
+	s.AddUpdateStatusRoute(h, "POST", "delivered", deliveredHandler)
+
+	failedHandler := h.buildStatusHandler("failed")
+	s.AddUpdateStatusRoute(h, "GET", "failed", failedHandler)
+	s.AddUpdateStatusRoute(h, "POST", "failed", failedHandler)
```
nicpottier: Can we check the log to see how they actually call us and only remap those methods?

handlers/shaqodoon/shaqodoon.go
```diff
@@ -1,6 +1,210 @@
...
+		return nil, err
+	}
+
+	return []courier.Msg{msg}, courier.WriteReceiveSuccess(w, r, msg)
+}
+
+type shaqodoonMessage struct {
+	From   string `name:"from"`
+	Sender string `name:"sender"`
+	Text   string `validate:"required" name:"text"`
```
nicpottier: Do they actually use both `from` and `sender`? (can check logs to find out)

handlers/shaqodoon/shaqodoon.go
norkans7: logs say no status report is ever sent to us

handlers/shaqodoon/shaqodoon.go
norkans7: logs show from only

PR #36, opened 2017-08-21
Telegram trigger conversation

norkans7:


backend.go
```diff
@@ -58,6 +58,9 @@ type Backend interface {
 	// StopMsgContact marks the contact for the passed in msg as stopped
 	StopMsgContact(Msg)
 
+	// TriggerNewConversation handle the trigger for new conversation for this contact
+	TriggerNewConversationForUrn(channel Channel, urn URN, name string) error
```
nicpottier: URN should be all caps in the method name.

backends/rapidpro/notifier.go
```diff
@@ -51,6 +51,14 @@ func (n *notifier) addStopContactNotification(contactID ContactID) {
 	n.notifications <- body
 }
 
+func (n *notifier) addTriggerNewConversation(contactID ContactID, channelID courier.ChannelID) {
```
nicpottier: How about `addNewConversationNotification`

norkans7:
This has been supported https://github.com/nyaruka/courier/blob/aecf212d950712b8cf71e30e7917a1e42048ab3e/handlers/telegram/telegram_test.go#L457

PR #37, opened 2017-08-21
Add librato logging

nicpottier:
I looked at a bunch of libraries to do this but they were all sucky in one way or another. Boo.

### DIFF
--- a/librato/librato.go
+++ b/librato/librato.go
@@ -1,0 +1,148 @@
+package librato
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/nyaruka/courier/utils"
+	"github.com/sirupsen/logrus"
+)
+
+// Default is our default librato collector
+var Default *Sender
+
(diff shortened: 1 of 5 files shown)


librato/librato.go
```diff
@@ -0,0 +1,148 @@
...
+
+// AddGauge can be used to add a new gauge to be sent to librato
+func (c *Sender) AddGauge(name string, value float64) {
+	// if no librato configured, return
+	if c == nil {
+		return
+	}
+
+	// our buffer is full, log an error but continue
+	if len(c.buffer) >= cap(c.buffer) {
```
rowanseymour: Does this need to be behind a mutex so two threads can't both get the all clear to to add to the buffer when it's only got room for one more item?

nicpottier: Maybe, but the idea is to avoid mutexes as much as possible with channels. We really should never reach this point as the buffer size should keep us ahead. If for some reason we do go beyond this, what will happen is the push unto the channel will block until our other goroutine flushes, so it isn't the end of the world.

PR #39, opened 2017-08-22
Various Tweaks

nicpottier:
* adds priority as top level item on Msg, lets Kannel send down priority appropriately
* optimize lookup as to whether things have been sent by stuffing it in a lua function
* clear sent status when we record an error status


backend.go
```diff
@@ -24,9 +24,6 @@ type Backend interface {
 	// NewIncomingMsg creates a new message from the given params
 	NewIncomingMsg(channel Channel, urn URN, text string) Msg
 
-	// NewOutgoingMsg creates a new outgoing message from the given params
-	NewOutgoingMsg(channel Channel, urn URN, text string) Msg
```
nicpottier: This isn't actually needed in the backend interface. (the test backend defines it but it is only used by it)

PR #40, opened 2017-09-04
changes to create and queue tasks for channel events

nicpottier:


backends/rapidpro/channel_event.go
```diff
@@ -0,0 +1,179 @@
...
+	"time"
+
+	null "gopkg.in/guregu/null.v3"
+
+	"github.com/nyaruka/courier"
+	"github.com/nyaruka/courier/utils"
+	"github.com/sirupsen/logrus"
+)
+
+// ChannelEventID is the type of our channel ids
```
rowanseymour: "channel event ids"

backends/rapidpro/channel_event.go
```diff
@@ -0,0 +1,179 @@
...
+type ChannelEventID struct {
```
rowanseymour: Why not `type ChannelEventID null.Int`?

PR #44, opened 2017-09-20
Use gocommon for URNs

rowanseymour:
See equivalent for goflow at https://github.com/nyaruka/goflow/pull/111


handlers/kannel/kannel.go
```diff
@@ -153,10 +153,8 @@ func (h *handler) SendMsg(msg courier.Msg) (courier.MsgStatus, error) {
 
 	// if we are meant to use national formatting (no country code) pull that out
 	if useNational {
-		parsed, err := phonenumbers.Parse(msg.URN().Path(), encodingDefault)
-		if err == nil {
-			form["to"] = []string{fmt.Sprintf("%d", parsed.GetNationalNumber())}
```
rowanseymour: This is a bit weird but I've kept the existing behavior. Because `GetNationalNumber` returns a uint64, it doesn't include any leading zeros, which are required for some countries.

rowanseymour: Also passing `encodingDefault` to `phonenumbers.Parse` instead of the channel country appears to be a bug?

nicpottier: This is the way it is by design, so no, adding the zero is not the right thing. I know that because I made that mistake the first time. The "0" is a dialing attribute of the number, it differentiates local dialing numbers on networks, it isn't actually the real "local number"

nicpottier: the `encodingDefault` is definitely wrong though, no doubt on that. Weird.

nicpottier:
Looks good apart from the national number bit.. wonder if `gocommon` is the right name to use here as that seems a bit broad as a package name. But then we'd have to name it `rapidprocommon` or something which seems kinda terrible. I guess `gocommon` is best option of those.

rowanseymour:
@nicpottier it keeps the existing behaviour for national numbers, but fixes the encodingDefault mistake, so I think we're good

PR #50, opened 2017-10-06
Chatbase Support

nicpottier:
If an org has chatbase enabled then this sends chatbase messages out.


nicpottier:
@teehamaral I know you are still ramping up on Golang but would be good to get your eyes on this as well. (tests incoming)

PR #53, opened 2017-11-13
Use callback domain when registering callbacks for sent messages

nicpottier:
Removed `Scheme` as a config as:
 1) I'm fine forcing https for callbacks (SSL everywhere!)
 2) and I thought it was a bit confusing as a config option as it felt like it was implying that we were serving via https (when really we are listening on a high port and SSL is handled upstream)

### DIFF
--- a/config/courier.go
+++ b/config/courier.go
@@ -10,7 +10,8 @@ type Courier struct {
 
 	SentryDSN string `default:""`
 
-	BaseURL  string `default:"https://localhost:8080"`
+	Scheme   string `default:"https"`
+	Domain   string `default:"localhost"`
 	Port     int    `default:"8080"`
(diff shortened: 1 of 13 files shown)


config/courier.go
```diff
@@ -6,38 +6,67 @@ import (
...
 	StatusPassword string `default:""`
 
+	// LogLevel controls the logging level courier uses
 	LogLevel string `default:"error"`
 
+	// IncludeChannels is the list of channels to enable, empty means include all
 	IncludeChannels []string
-	ExcludeChannels []string
 
-	IgnoreTwilioStatus bool `default:"false"`
```
rowanseymour: This brings back fond memories...

PR #54, opened 2017-11-15
Infobip updates

norkans7:


handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+// NewHandler returns a new Infobip handler
+func NewHandler() courier.ChannelHandler {
+	return &handler{handlers.NewBaseHandler(courier.ChannelType("IB"), "Infobip")}
+}
+
+// Initialize is called by the engine once everything is loaded
+func (h *handler) Initialize(s courier.Server) error {
+	h.SetServer(s)
+	err := s.AddReceiveMsgRoute(h, "POST", "receive", h.ReceiveMessage)
+	return err
```
nicpottier: can be one line here

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if ie.MessageCount == 0 {
+		return nil, courier.WriteIgnored(w, r, "Ignoring request, no message")
+	}
+
+	msgs := []courier.Msg{}
+	if len(ie.Results) > 0 {
```
nicpottier: Don't need this if

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+		for index, infobipMessage := range ie.Results {
+			fmt.Println(index)
```
nicpottier: Get rid of print

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+			messageID := infobipMessage.MessageID
+			text := infobipMessage.Text
+			//receiver := infobipMessage.To
```
nicpottier: Remove if unused

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+			if dateString != "" {
+				date, err = time.Parse("2006-01-02T15:04:05.999999999-0700", dateString)
+				if err != nil {
+					date = time.Now()
+				}
+			}
+			date = date.UTC()
+
+			// create our URN
+			urn := courier.NewTelURNForChannel(sender, channel)
```
nicpottier: Would just not declare a local variable for sender above and do this inline instead

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+			text := infobipMessage.Text
+			//receiver := infobipMessage.To
+			sender := infobipMessage.From
+			dateString := infobipMessage.ReceivedAt
+
+			date := time.Now()
+			if dateString != "" {
+				date, err = time.Parse("2006-01-02T15:04:05.999999999-0700", dateString)
+				if err != nil {
+					date = time.Now()
```
nicpottier: Should probably return an error if passed an invalid date unless we know this is something they do

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+				}
+			}
+			date = date.UTC()
```
nicpottier: No need to normalize to UTC, might as well use the timezone they defined

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+
+			// create our URN
+			urn := courier.NewTelURNForChannel(sender, channel)
+
+			// build our infobipMessage
+			msg := h.Backend().NewIncomingMsg(channel, urn, text).WithReceivedOn(date).WithExternalID(messageID)
+
+			// and write it
+			h.Backend().WriteMsg(msg)
```
nicpottier: Should be handling error here

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+	err := s.AddReceiveMsgRoute(h, "POST", "receive", h.ReceiveMessage)
+	return err
+}
+
+// ReceiveMessage is our HTTP handler function for incoming messages
+func (h *handler) ReceiveMessage(channel courier.Channel, w http.ResponseWriter, r *http.Request) ([]courier.Msg, error) {
+	ie := &infobipEnvelope{}
+	err := handlers.DecodeAndValidateJSON(ie, r)
+	if err != nil {
+		return nil, err
```
nicpottier: Use `courier.WriteError()` here as this is a validation error

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+			// create our URN
+			urn := courier.NewTelURNForChannel(sender, channel)
+
+			// build our infobipMessage
+			msg := h.Backend().NewIncomingMsg(channel, urn, text).WithReceivedOn(date).WithExternalID(messageID)
+
+			// and write it
+			h.Backend().WriteMsg(msg)
+			msgs = append(msgs, msg)
+			courier.WriteReceiveSuccess(w, r, msg)
```
nicpottier: Shouldn't write this for every message if they can send more than one

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+		}
+	}
+
+	if len(msgs) == 0 {
+		return nil, errors.New("No message found")
```
nicpottier: Use `writeError` here as this is a validation error

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+	}
+
+	return msgs, nil
+}
+
+type infobipMessage struct {
+	MessageID  string `json:"messageId"`
+	From       string `json:"from" validate:"required"`
+	To         string `json:"to" validate:"required"`
+	Text       string `json:"text" validate:"required"`
```
nicpottier: We probably don't want `Text` to be required if we want to allow empty messages. (which definitely occurs)

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+// 		  "currency": "EUR"
+// 		},
+// 		"callbackData": "callbackData"
+// 	  }
+// 	],
+// 	"messageCount": 1,
+// 	"pendingMessageCount": 0
+// }
+type infobipEnvelope struct {
+	PendingMessageCount int64            `json:"pendingMessageCount"`
```
nicpottier: can just use int here

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+		}
+		msgs = append(msgs, msg)
+
+	}
+
+	if len(msgs) == 0 {
+		return nil, courier.WriteError(w, r, errors.New("no message found"))
+	}
+
+	return []courier.ReceiveEvent{msgs[0]}, courier.WriteMsgSuccess(w, r, msgs[0])
```
nicpottier: Should return all the messages here, not just the first. We can tweak `WriteMsgSuccess` to take the list of messages as well.

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+		err = h.Backend().WriteMsg(msg)
+		if err != nil {
+			return nil, courier.WriteError(w, r, err)
+		}
+		msgs = append(msgs, msg)
+
+	}
+
+	if len(msgs) == 0 {
+		return nil, courier.WriteError(w, r, errors.New("no message found"))
```
nicpottier: In what cases would this occur? Why is this an error but having no list an ignore?

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,124 @@
...
+		// create our URN
+		urn := urns.NewTelURNForCountry(infobipMessage.From, channel.Country())
+
+		// build our infobipMessage
+		msg := h.Backend().NewIncomingMsg(channel, urn, text).WithReceivedOn(date).WithExternalID(messageID)
+
+		// and write it
+		err = h.Backend().WriteMsg(msg)
+		if err != nil {
+			return nil, courier.WriteError(w, r, err)
```
nicpottier: In this case, this is a courier error, not a validation error, so we should return err. (that will get logged to sentry and the caller will take care of returning a formatted JSON error)

norkans7: Okay I get that now

handlers/infobip/infobip.go
norkans7: If we had a list of messages without text

handlers/infobip/infobip.go
norkans7: Okay

handlers/infobip/infobip.go
norkans7: also we have a list some could be missing the text and others valid to create a message, currently we create as much msgs as we can and ignore the invalid dicts

nicpottier: Ya, understood, but why is it an error here but an ignore on line 89? If we get a payload that has no messages to deal with then we should be consistent in either returning an error or just saying that's fine and we ignore it.

norkans7: Because we had the count greater than on and no text for the message like https://github.com/nyaruka/courier/pull/54/files#diff-abfdf7a6298b5a6cbf9404b33e084a46R68 so we ignored that in 99

if not message is created for the request then we return an error here

nicpottier: But why? Why is that case an error (empty text) but having no messages an ignore?

norkans7: Okay I can ignore both cases then

norkans7:
@nicpottier I added full support for Infobip now, please review once again

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,248 @@
...
+
+var infobipStatusMapping = map[string]courier.MsgStatusValue{
+	"PENDING":       courier.MsgSent,
+	"EXPIRED":       courier.MsgSent,
+	"DELIVERED":     courier.MsgDelivered,
+	"REJECTED":      courier.MsgFailed,
+	"UNDELIVERABLE": courier.MsgFailed,
+}
+
+type ibStatusEnvelop struct {
```
nicpottier: `envelope` not `envelop`

nicpottier: It'd be nice to have an example payload here as a comment just as we do for received messages below.

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,248 @@
...
+	authHeader := "Basic " + encodedCreds
+
+	callbackDomain := msg.Channel().CallbackDomain(h.Server().Config().Domain)
+	statusURL := fmt.Sprintf("https://%s%s%s/delivered", callbackDomain, "/c/ib/", msg.Channel().UUID())
+
+	ibMsg := ibOutgoingEnvelop{
+		Messages: []ibOutgoingMessage{
+			ibOutgoingMessage{
+				From: msg.Channel().Address(),
+				Destinations: []ibDestination{ibDestination{
```
nicpottier: I think put `ibDestination` on its own line

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,248 @@
...
+				Text:               courier.GetTextAndAttachments(msg),
+				NotifyContentType:  "application/json",
+				IntermediateReport: true,
+				NotifyURL:          statusURL,
+			},
+		},
+	}
+
+	requestBody := new(bytes.Buffer)
+	json.NewEncoder(requestBody).Encode(ibMsg)
```
nicpottier: Should be checking err here.

handlers/infobip/infobip.go
```diff
@@ -1,3 +1,248 @@
...
+				}},
+				Text:               courier.GetTextAndAttachments(msg),
+				NotifyContentType:  "application/json",
+				IntermediateReport: true,
+				NotifyURL:          statusURL,
+			},
+		},
+	}
+
+	requestBody := new(bytes.Buffer)
```
nicpottier: Dont need new here, just do:
```
requestBody := bytes.Buffer{}
```

utils/misc.go
```diff
@@ -86,3 +87,12 @@ func CleanString(s string) string {
...
 }
+
+// EncodeBase64 encodes the list of strings with Linux base64
+func EncodeBase64(strList []string) string {
+	var strBuffer bytes.Buffer
+	for _, l := range strList {
+		strBuffer.WriteString(l)
+	}
+	return base64.StdEncoding.EncodeToString([]byte(strBuffer.String()))
+}
```
nicpottier: Do we need this? Why not just:

`base64.StdEncoding.EncodeToString([]byte(strings.Join(strList, ""))`

nicpottier:
Last thing, since these are loaded dynamically, you want to import this in `cmd/courier/main.go`

Change that and I'll merge and ship this to rapidpro and we can see how it works for receiving messages

PR #55, opened 2017-11-16
Update write resonse helpers

norkans7:


responses.go
```diff
@@ -44,30 +44,43 @@ func WriteChannelEventSuccess(w http.ResponseWriter, r *http.Request, event Chan
...
+				msg.Channel().UUID(),
+				msg.UUID(),
+				msg.Text(),
+				msg.URN(),
+				msg.Attachments(),
+				msg.ExternalID(),
+				msg.ReceivedOn(),
+			})
+	}
+	return writeData(w, http.StatusOK, "Message Accepted", data)
```
nicpottier: Feels slightly weird to me that now data can be a JSON list or a JSON dictionary. I feel like it might be friendlier to callers if we made sure data was always a dictionary. So in this case we would want to wrap the `msgReceiveData` in another struct where that was a key like `msgs`.

responses.go
```diff
@@ -44,30 +44,43 @@ func WriteChannelEventSuccess(w http.ResponseWriter, r *http.Request, event Chan
...
+			data,
+			statusData{
+				status.ChannelUUID(),
+				status.Status(),
+				status.ID(),
+				status.ExternalID(),
+			})
+	}
+
+	return writeData(w, http.StatusOK, "Status Update Accepted", data)
```
nicpottier: Here would be `statuses` in `data` (so the list within the a struct)

norkans7:
Please re-review @nicpottier

PR #56, opened 2017-11-17
FB Channel

nicpottier:
@norkans7 this isn't done yet but would be a good next channel to port over. (viber being the next)


nicpottier:
Actually looking at this it is pretty far off, so I'll take care of it (there's some weirdness around FB referral ids which needs working out).

@norkans7 why don't you do viber next?

johncordeiro:
What's happening on facebook referral ids? Maybe we can help with that.

nicpottier:
We're working on it, thanks though!

handlers/facebook/facebook.go
```diff
@@ -1,6 +1,386 @@
...
+func init() {
+	courier.RegisterHandler(NewHandler())
+}
+
+type handler struct {
+	handlers.BaseHandler
+}
+
+const (
+	maxLength = 640 // Facebook API says 640 is max for the body
```
norkans7: this const can join the other on top

PR #59, opened 2017-11-23
Viber public handlers

norkans7:


handlers/viber/viber.go
```diff
@@ -10,3 +27,280 @@ POST /handlers/viber_public/uuid?sig=sig
...
+	confAuth := msg.Channel().ConfigForKey(courier.ConfigAuthToken, "")
+	authToken, isStr := confAuth.(string)
+	if !isStr || authToken == "" {
+		return nil, fmt.Errorf("invalid auth token config")
+	}
+
+	viberMsg := viberOutgoingMessage{
+		AuthToken:    authToken,
+		Receiver:     msg.URN().Path(),
+		Text:         courier.GetTextAndAttachments(msg),
```
nicpottier: Whats the max message length? Does viber not support attachments?

norkans7: I added the max length, for attachement we can easily support picture as attachment but for video we need to specify the size of the attachment https://developers.viber.com/docs/api/rest-bot-api/#video-message 
I think we need first a way to get that do we fetch first the mediaURL to get the size?

handlers/viber/viber.go
```diff
@@ -10,3 +31,386 @@ POST /handlers/viber_public/uuid?sig=sig
...
+}
+
+// see https://developers.viber.com/docs/api/rest-bot-api/#callbacks
+func (h *handler) validateSignature(channel courier.Channel, r *http.Request) error {
+	actual := r.Header.Get(viberSignatureHeader)
+	if actual == "" {
+		return fmt.Errorf("missing request signature")
+	}
+
+	confAuth := channel.ConfigForKey(courier.ConfigAuthToken, "")
```
nicpottier: `channel.StringConfigForKey` will save you a step.

handlers/viber/viber.go
```diff
@@ -10,3 +31,386 @@ POST /handlers/viber_public/uuid?sig=sig
...
+	authToken, isStr := confAuth.(string)
+	if !isStr || authToken == "" {
+		return fmt.Errorf("invalid or missing auth token in config")
+	}
+
+	// read our body
+	body, err := ioutil.ReadAll(r.Body)
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	expected, err := viberCalculateSignature(authToken, body)
```
nicpottier: no need for the `viber` prefix, it is a private method

handlers/viber/viber.go
```diff
@@ -10,3 +31,386 @@ POST /handlers/viber_public/uuid?sig=sig
...
+		return fmt.Errorf("invalid request signature")
+	}
+
+	return nil
+}
+
+func viberCalculateSignature(authToken string, contents []byte) ([]byte, error) {
+
+	var buffer bytes.Buffer
+	buffer.Write(contents)
```
nicpottier: what purpose does buffer serve? can't we just write the contents directly below?

handlers/viber/viber.go
```diff
@@ -10,3 +31,386 @@ POST /handlers/viber_public/uuid?sig=sig
...
+	AuthToken    string            `json:"auth_token"`
+	Receiver     string            `json:"receiver"`
+	Text         string            `json:"text"`
+	Type         string            `json:"type"`
+	TrackingData string            `json:"tracking_data"`
+	Sender       map[string]string `json:"sender,omitempty"`
+	Media        string            `json:"media,omitempty"`
+	Size         int               `json:"size,omitempty"`
+	FileName     string            `json:"file_name,omitempty"`
+	Keyboard     string            `json:"keyboard,omitempty"`
```
nicpottier: This should be a pointer to `viberKeyboard`, not a string. Right now it is going to turn into a double encoded JSON blob

utils/http.go
```diff
@@ -16,14 +16,15 @@ type RequestResponseStatus string
...
-	Body       []byte
-	Elapsed    time.Duration
+	Method        string
+	URL           string
+	Status        RequestResponseStatus
+	StatusCode    int
+	Request       string
+	Response      string
+	Body          []byte
+	ContentLength string
```
nicpottier: Feels like this should be an int (perhaps with a value of -1 if unset)

handlers/viber/viber.go
norkans7: that way the omitempty is not effective and there is keyboard: {} in the JSON

channel_event.go
```diff
@@ -13,6 +13,7 @@ type ChannelEventType string
 const (
 	NewConversation ChannelEventType = "new_conversation"
 	Referral        ChannelEventType = "referral"
+	StopContact     ChannelEventType = "stop_contact"
```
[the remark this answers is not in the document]
norkans7: added in https://github.com/nyaruka/rapidpro/pull/1688

utils/http.go
norkans7: not sure how to set -1 in that case.. should I change RequestResponse{} to always be RequestResponse{ContentLength: -1}?

handlers/viber/viber.go
nicpottier: You make it a pointer to a `viberKeyboard`, omit empty will leave it out then.

IE: 

`Keyboard *viberKeyboard`

utils/http.go
nicpottier: Yes, and should stay -1 if there is no ContentLength header

handlers/viber/viber_test.go
```diff
@@ -0,0 +1,412 @@
...
+			}
+		}
+		testCases[c].RequestBody = strings.Replace(testCases[c].RequestBody, "{{ SERVER_URL }}", server.URL, -1)
+	}
+
+	return server
+}
+
+var defaultSendTestCases = []ChannelSendTestCase{
+	{Label: "Plain Send",
```
nicpottier: Should add a QuickReplies case.

handlers/viber/viber.go
norkans7: I have tried this and I keep getting {} in the JSON

norkans7: never mind, this is now updated

PR #60, opened 2017-11-23
Support receiving message for Viber

norkans7:


norkans7:
Two things still left TO DO adding signature validation
and adding the handle for StopContact channelEvent handle on rapidPro side to be able to stop and contact by using its URN.

norkans7:
Two things still left TO DO adding signature validation
and adding the handle for StopContact channelEvent handle on rapidPro side to be able to stop and contact by using its URN.

PR #62, opened 2017-12-01
Add basic handling for whatsapp

nicpottier:
No media handling yet this is just basic messaging.


handlers/whatsapp/whatsapp.go
```diff
@@ -0,0 +1,282 @@
...
+type waReceive struct {
+	Payload struct {
+		From      string `json:"from" validate:"required"`
+		MessageID string `json:"message_id" validate:"required"`
+		Timestamp string `json:"timestamp" validate:"required"`
+		Message   struct {
+			HasMedia  bool    `json:"has_media"`
+			Address   string  `json:"address"`
+			Latitude  float64 `json:"latitude"`
+			Longitude float64 `json:"longitude"`
```
norkans7: It seems there are missing fields here for the media case but I saw the TODOs

nicpottier: Media is super weird in that it will actually come in a totally different request.. but no, these are all the fields present / possible. But ya, not implemented fully yet.

PR #68, opened 2017-12-16
pass context down to backend to support timeouts

nicpottier:
Fixes problem where courier falls over when DB goes inaccessible.


nicpottier:
Fixes https://github.com/nyaruka/courier/issues/58

PR #72, opened 2017-12-21
Redis auth

jlev:
Send authentication command if the Redis host requires a password to connect

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -388,12 +388,14 @@ func (b *backend) Start() error {
 			}
 
 			// send auth if required
-			pass, authRequired := redisURL.User.Password()
-			if authRequired {
-				if _, err := conn.Do("AUTH", pass); err != nil {
-			      conn.Close()
-			      return nil, err
-			    }
+			if redisURL.User != nil {
+				pass, authRequired := redisURL.User.Password()
+				if authRequired {
+					if _, err := conn.Do("AUTH", pass); err != nil {
+				      conn.Close()
+				      return nil, err
+				    }
+				}
 			}
 
 			// switch to the right DB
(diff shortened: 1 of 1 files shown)


backends/rapidpro/backend.go
```diff
@@ -370,11 +370,17 @@ func (b *backend) Start() error {
 	}
 
 	// parse and test our redis config
+	log.Info("redis config")
```
nicpottier: can remove this

backends/rapidpro/backend.go
```diff
@@ -370,11 +370,17 @@ func (b *backend) Start() error {
...
 	redisURL, err := url.Parse(b.config.Redis)
 	if err != nil {
 		return fmt.Errorf("unable to parse Redis URL '%s': %s", b.config.Redis, err)
 	}
 
+	logrus.WithFields(logrus.Fields{
+		"host":  redisURL.Host,
+		"auth":  redisURL.User,
+	}).Info("parsed redis config")
```
nicpottier: remove this

backends/rapidpro/backend.go
```diff
@@ -387,6 +393,15 @@ func (b *backend) Start() error {
 				return nil, err
 			}
 
+			// send auth if required
+			pass, auth_required := redisURL.User.Password()
```
nicpottier: GoLang variables are camelcase, so `authRequired` here would probably be best.

nicpottier:
Looks like build is failing because User can be `nil`

nicpottier:
Sorry to be a pain but this needs to be run through gofmt or the indenting needs to be fixed up. :)

nicpottier:
woot, thanks!

PR #74, opened 2018-01-10
Add response and log method for ignoring status messages for messages…

nicpottier:
… we don't know about.

Just standardizing these.. we were filling up sentry with errors on the Nexmo side due to them splitting message ids and giving us status for each segment. This just changes us to have a shared method to call for the response and message.

### DIFF
--- a/handlers/facebook/facebook.go
+++ b/handlers/facebook/facebook.go
@@ -333,7 +333,7 @@ func (h *handler) Receive(ctx context.Context, channel courier.Channel, w http.R
 
 				// we don't know about this message, just tell them we ignored it
 				if err == courier.ErrMsgNotFound {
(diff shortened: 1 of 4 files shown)


handlers/facebook/facebook.go
```diff
@@ -333,7 +333,7 @@ func (h *handler) Receive(ctx context.Context, channel courier.Channel, w http.R
 
 				// we don't know about this message, just tell them we ignored it
 				if err == courier.ErrMsgNotFound {
-					data = append(data, courier.NewInfoData("unknown message, ignoring"))
+					data = append(data, courier.NewInfoData("message not found, ignored"))
```
norkans7: may be use the const here too

nicpottier: its private which I think I want to keep for the time being

PR #89, opened 2018-01-24
WIP handling Jiochat requests

norkans7:


server.go
```diff
@@ -329,7 +329,10 @@ func (s *server) AddHandlerRoute(handler ChannelHandler, method string, action s
 	method = strings.ToLower(method)
 	channelType := strings.ToLower(string(handler.ChannelType()))
 
-	path := fmt.Sprintf("/%s/{uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}}/%s", channelType, action)
+	path := fmt.Sprintf("/%s/{uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}}", channelType)
```
norkans7: The callback URL confifured looks like `/c/jc/abcd-abcd-abcd-abcd/` 

Here I make sure we support those for `""` action passed to the AddHandlerRoute

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return []courier.Event{msg}, courier.WriteMsgSuccess(ctx, w, r, []courier.Msg{msg})
+}
+
+var mediaDownloadURL = "https://channels.jiochat.com/media/download.action"
+
+func resolveFileID(mediaID string) string {
```
norkans7: `resolveFileID` here is just building the URL however `msg.go` won't be able to `downloadAndSavetoS3` without access to jiochat channel access token, We need to discuss the approach to notify the authentication to msg.go

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+		StatusCode = 200
+	}
+
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(StatusCode)
+	_, err = fmt.Fprint(w, ResponseText)
+	return nil, err
+}
+
+type jiochatMsgRequest struct {
```
nicpottier: I've been calling these `moMsg`

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+	Content      string `json:"Content"`
+	MediaID      string `json:"MediaId"`
+}
+
+// ReceiveMessage is our HTTP handler function for incoming messages
+func (h *handler) ReceiveMessage(ctx context.Context, channel courier.Channel, w http.ResponseWriter, r *http.Request) ([]courier.Event, error) {
+	jcRequest := &jiochatMsgRequest{}
+	err := handlers.DecodeAndValidateJSON(jcRequest, r)
+	if err != nil {
+		return nil, err
```
nicpottier: write error request

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	if jcRequest.FromUsername == "" || jcRequest.MsgType == "" || (jcRequest.MsgID == 0 && jcRequest.Event == "") {
```
nicpottier: Some of these can be in the validate tags and then will be taken care of above

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	date := time.Unix(jcRequest.CreateTime, 0).UTC()
+
+	urn := urns.NewURNFromParts(urns.JiochatScheme, jcRequest.FromUsername, "")
+
+	if jcRequest.MsgType == "event" && jcRequest.Event == "subscribe" {
+
+		// build the channel event
+		channelEvent := h.Backend().NewChannelEvent(channel, courier.Follow, urn)
```
nicpottier: what is the difference between `follow` and `new_conversation`?

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+
+var mediaDownloadURL = "https://channels.jiochat.com/media/download.action"
+
+func resolveFileID(mediaID string) string {
+	form := url.Values{
+		"media_id": []string{mediaID},
+	}
+	encodedForm := form.Encode()
+	mediaURL := fmt.Sprintf("%s?%s", mediaDownloadURL, encodedForm)
+	return mediaURL
```
nicpottier: safer way to do this:
```go
mediaURL, _ := url.Parse(sendURL)
mediaURL.RawQuery = form.Encode()
```

the use `mediaURL.String()`

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func getAccessToken(channel courier.Channel) string {
+	return ""
```
nicpottier: hrmm?

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,197 @@
...
+func (h *handler) DescribeURN(ctx context.Context, channel courier.Channel, urn urns.URN) (map[string]string, error) {
+	accessToken := getAccessToken(channel)
+
+	_, path, _ := urn.ToParts()
+
+	form := url.Values{
+		"openid": []string{path},
+	}
+	encodedForm := form.Encode()
+	userDetailsURL = fmt.Sprintf("%s?%s", userDetailsURL, encodedForm)
```
nicpottier: See above

backends/rapidpro/msg.go
```diff
@@ -181,17 +183,37 @@ func readMsgFromDB(b *backend, id courier.MsgID) (*DBMsg, error) {
...
 	}
 
-	// first fetch our media
-	req, err := http.NewRequest("GET", mediaURL, nil)
-	if err != nil {
-		return "", err
+	var req *http.Request
+	handler := courier.GetHandler(channel.ChannelType())
+	if handler != nil {
+		mediaDownloadRequestBuilder, isMediaDownloadRequestBuilder := handler.(courier.MediaDownloadRequestBuilder)
```
nicpottier: I like this approach, how about some shorter variable names? Go's style says use the shortest possible names.

So just using `builder` and `isBuilder` here would be better for example.

norkans7:
This is not ready, It still needs sending  and the getAccessToken method that need to use Redis to get the token

norkans7:
@nicpottier please review once again

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+
+	jsonBody, err := json.Marshal(refreshTokenData)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, tokenURL.String(), bytes.NewReader(jsonBody))
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+	rr, err := utils.MakeHTTPRequest(req)
```
nicpottier: Should be checking error here

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+
+var jiochatURL = "https://channels.jiochat.com"
+
+func init() {
+	courier.RegisterHandler(newHandler(&jiochatClient{}))
+}
+
+type handler struct {
+	handlers.BaseHandler
+	jiochatClient jiochatAccess
```
nicpottier: I don't understand the purpose of this client / interface? Why have it?

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+
+	err = h.Backend().WriteMsg(ctx, msg)
+	if err != nil {
+		return nil, err
+	}
+
+	return []courier.Event{msg}, courier.WriteMsgSuccess(ctx, w, r, []courier.Msg{msg})
+}
+
+func resolveMediaID(mediaID string) string {
```
nicpottier: This isn't really resolving, just building a media URL, so maybe something more like `buildMediaURL` would fit better.

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+func (jCli *jiochatClient) getAccessToken(h *handler, channel courier.Channel) string {
+	rc := h.Backend().RedisPool().Get()
+	defer rc.Close()
+	cacheKey := fmt.Sprintf("jiochat_channel_access_token:%s", channel.UUID().String())
+	accessToken, _ := redis.String(rc.Do("GET", cacheKey))
+	return accessToken
+}
+
+func (h *handler) findAccessToken(channel courier.Channel) string {
+	return h.jiochatClient.getAccessToken(h, channel)
```
nicpottier: Single line methods like this are anti-patterns, why even have it?

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (h *handler) doFetchAccessToken(channel courier.Channel) error {
+	return h.jiochatClient.fetchAccessToken(h, channel)
```
nicpottier: Same here.

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+type mtMsg struct {
+	MsgType string `json:"msgtype"`
+	ToUser  string `json:"touser"`
```
nicpottier: no underscore? it is really `touser`?

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+		Content string `json:"content"`
+	} `json:"text"`
+}
+
+// SendMsg sends the passed in message, returning any error
+func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStatus, error) {
+	accessToken := h.findAccessToken(msg.Channel())
+
+	status := h.Backend().NewMsgStatusForID(msg.Channel(), msg.ID(), courier.MsgErrored)
+	parts := handlers.SplitMsg(handlers.GetTextAndAttachments(msg), maxMsgLength)
```
nicpottier: No MT media support?

handlers/jiochat/jiochat.go
```diff
@@ -0,0 +1,323 @@
...
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", accessToken))
+	if err != nil {
+		return nil, err
+	}
+
+	rr, err := utils.MakeHTTPRequest(req)
+	if err != nil {
+		return nil, fmt.Errorf("unable to look up contact data:%s\n%s", err, rr.Response)
+	}
+	nickname, _ := jsonparser.GetString(rr.Body, "nickname")
```
nicpottier: Worth checking the error here otherwise we are passing back an empty name.

handlers/jiochat/jiochat_test.go
```diff
@@ -0,0 +1,366 @@
...
+	err         error
+}
+
+func (ja *jiochatAccessMock) fetchAccessToken(h *handler, channel courier.Channel) error {
+	return ja.err
+}
+
+func (ja *jiochatAccessMock) getAccessToken(h *handler, channel courier.Channel) string {
+	return ja.accessToken
+}
```
nicpottier: I would rather see the mock server satisfy these requests so this is all tested, as is we aren't really testing any of the access token stuff.

handlers/jiochat/jiochat.go
norkans7: yep

handlers/jiochat/jiochat.go
norkans7: That requires to first upload the file and get an id, we did not support that on python I think we make the current behavior work and add the new support in a different PR

handlers/jiochat/jiochat.go
norkans7: the interface allow me to mock the redis and be able to test the header is properly tested

handlers/jiochat/jiochat.go
norkans7: this is influenced by the interface to be able to test

handlers/jiochat/jiochat.go
norkans7: okay

handlers/jiochat/jiochat.go
nicpottier: Ok, I'm fine with that being another PR.

handlers/jiochat/jiochat.go
nicpottier: Why do you need that vs a mock server to satisfy these? (and a configurable URL that they hit?)

handlers/jiochat/jiochat_test.go
norkans7: I'll check more on this, that is the issue I had I was not able to test the header was set correctly and used the interface so I can mock it in the tests, I try checking on this suggestion then

handlers/jiochat/jiochat.go
norkans7: I am unable to set values on Redis on the mock server

test.go
```diff
@@ -31,13 +33,42 @@ type MockBackend struct {
...
+	}
+	conn := redisPool.Get()
+	defer conn.Close()
+
+	_, err := conn.Do("FLUSHDB")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	_, err = conn.Do("Set", "jiochat_channel_access_token:8eb23e93-5ecb-45ba-b726-3b064e0c56ab", "ACCESS_TOKEN")
```
nicpottier: This should not be in the global backend

norkans7:
@nicpottier ready for one more review

PR #93, opened 2018-01-26
add handler for m3tech

nicpottier:


handlers/base.go
```diff
@@ -238,3 +241,48 @@ func SplitMsg(text string, max int) []string {
...
+		if err != nil {
+			return nil, err
+		}
+
+		return []courier.Event{msg}, courier.WriteMsgSuccess(ctx, w, r, []courier.Msg{msg})
+	}
+}
+
+// GetTextAndAttachments returns both the text of our message as well as any attachments, newline delimited
+func GetTextAndAttachments(m courier.Msg) string {
```
nicpottier: Moved this and the one below to handlers since thats the only place they are used.

PR #111, opened 2018-02-14
further tweaks to mtarget

nicpottier:
Allow sending unicode, deal with multiparts that come in separately.

### DIFF
--- a/handlers/mtarget/mtarget.go
+++ b/handlers/mtarget/mtarget.go
@@ -5,9 +5,12 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/buger/jsonparser"
+	"github.com/garyburd/redigo/redis"
 	"github.com/nyaruka/courier"
 	"github.com/nyaruka/courier/handlers"
 	"github.com/nyaruka/courier/utils"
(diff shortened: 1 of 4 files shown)


handlers/mtarget/mtarget.go
```diff
@@ -65,6 +68,65 @@ func (h *handler) receiveMsg(ctx context.Context, c courier.Channel, w http.Resp
...
+			keys[i] = fmt.Sprintf("%d", i)
+		}
+
+		segments, err := redis.Strings(rc.Do("HMGET", keys...))
+		if err != nil {
+			return nil, err
+		}
+
+		// join our segments in our text
+		text = strings.Join(segments, "")
```
norkans7: I think we need to join using a space

nicpottier: No, these are multipart messages, they aren't necessarily split on a space.

handlers/mtarget/mtarget.go
```diff
@@ -65,6 +68,65 @@ func (h *handler) receiveMsg(ctx context.Context, c courier.Channel, w http.Resp
...
+		}
+
+		rc := h.Backend().RedisPool().Get()
+		defer rc.Close()
+
+		// first things first, populate the new part we just received
+		mapKey := fmt.Sprintf("%s:%s", c.UUID().String(), longID)
+		rc.Send("MULTI")
+		rc.Send("HSET", mapKey, longRef, text)
+		rc.Send("EXPIRE", mapKey, 300)
```
norkans7: is 5min enough to wait all the parts?

nicpottier: I think so, we can tweak if not, but there shouldn't be that big a delay.

PR #114, opened 2018-02-15
Line handlers

### LINKED ISSUE #96: Add Line Channel

norkans7:
Fixes #96


handlers/line/line.go
```diff
@@ -1,3 +1,178 @@
...
+
+	if len(msgs) == 0 {
+		if len(lineRequest.Events) > 0 {
+			return nil, courier.WriteAndLogRequestIgnored(ctx, w, r, channel, "ignoring request, no message")
+		}
+		return nil, courier.WriteAndLogRequestError(ctx, w, r, channel, fmt.Errorf("missing message, source or type in the event"))
+
+	}
+
+	return []courier.Event{msgs[0]}, courier.WriteMsgSuccess(ctx, w, r, msgs)
```
nicpottier: Should be returning all msgs, not just [0]

handlers/line/line.go
```diff
@@ -1,3 +1,178 @@
...
+// 		},
+// 		"message": {
+// 		  "id": "325708",
+// 		  "type": "text",
+// 		  "text": "Hello, world"
+// 		}
+// 	  },
+// 	  {
+// 		"replyToken": "nHuyWiB7yP5Zw52FIkcQobQuGDXCTA",
+// 		"type": "follow",
```
nicpottier: Do we not support follows?

handlers/line/line.go
```diff
@@ -1,3 +1,178 @@
...
+			return nil, err
+		}
+		msgs = append(msgs, msg)
+	}
+
+	if len(msgs) == 0 {
+		if len(lineRequest.Events) > 0 {
+			return nil, courier.WriteAndLogRequestIgnored(ctx, w, r, channel, "ignoring request, no message")
+		}
+		return nil, courier.WriteAndLogRequestError(ctx, w, r, channel, fmt.Errorf("missing message, source or type in the event"))
```
nicpottier: Not sure I understand this case..  is this really an error?

handlers/line/line.go
```diff
@@ -1,3 +1,178 @@
...
+
+		rr, err := utils.MakeHTTPRequest(req)
+		// record our status and log
+		log := courier.NewChannelLogFromRR("Message Sent", msg.Channel(), msg.ID(), rr).WithError("Message Send Error", err)
+		status.AddLog(log)
+
+		if err != nil {
+			return status, err
+		}
+		status.SetStatus(courier.MsgWired)
```
nicpottier: Any external id to parse?

handlers/line/line.go
```diff
@@ -1,3 +1,178 @@
...
+	"github.com/nyaruka/gocommon/urns"
+	"net/http"
+	"time"
+
+	"github.com/nyaruka/courier"
+	"github.com/nyaruka/courier/handlers"
+)
+
+var sendURL = "https://api.line.me/v2/bot/message/push"
+var maxMsgLength = 1600
```
nicpottier: Do we know what this is exactly from their API docs?

norkans7: I see it is 2000 on the docs, I update that

handlers/line/line.go
norkans7: the docs says they just return a 200 and and empty JSON

handlers/line/line.go
norkans7: If the events list was empty we return an error, otherwise the event were just ignored by us

handlers/line/line.go
norkans7: No

handlers/line/line.go
nicpottier: Seems if the list is empty we should just ignore it too, if its valid JSON that's just fine.

handlers/line/line.go
nicpottier: Any reason why not?

nicpottier: I guess if we don't support it in the past that's ok to leave it off for now.

norkans7: right

PR #118, opened 2018-02-20
Implement firebase

nicpottier:
Slightly more complicated as we need to introduce the idea of contact creation as well as auth on URN.

### DIFF
--- a/handlers/firebase/firebase.go
+++ b/handlers/firebase/firebase.go
@@ -1,3 +1,215 @@
 package firebase
 
-/* no logs */
+import (
+	"bytes"
+	"context"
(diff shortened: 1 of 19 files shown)


handlers/firebase/firebase.go
```diff
@@ -1,3 +1,215 @@
...
+)
+
+const (
+	configTitle        = "FCM_TITLE"
+	configNotification = "FCM_NOTIFICATION"
+	configKey          = "FCM_KEY"
+)
+
+var sendURL = "https://fcm.googleapis.com/fcm/send"
+
```
norkans7: no maxMsgLength?

PR #119, opened 2018-02-21
Chikka handlers

norkans7:


handlers/chikka/chikka.go
```diff
@@ -1,3 +1,148 @@
...
+
+		msgStatus, found := statusMapping[ckRequest.Status]
+		if !found {
+			return nil, courier.WriteAndLogRequestError(ctx, w, r, channel, fmt.Errorf(`unknown status '%s', must be either 'SENT' or 'FAILED'`, ckRequest.Status))
+		}
+
+		// write our status
+		status := h.Backend().NewMsgStatusForID(channel, courier.NewMsgID(ckRequest.MessageID), msgStatus)
+		err = h.Backend().WriteMsgStatus(ctx, status)
+		if err != nil {
```
nicpottier: Should also deal with the case where we get an error about the msg not existing.

handlers/chikka/chikka.go
```diff
@@ -1,3 +1,148 @@
...
+			return nil, err
+		}
+
+		return []courier.Event{status}, courier.WriteStatusSuccess(ctx, w, r, []courier.MsgStatus{status})
+
+	} else if ckRequest.MessageType == "incoming" {
+
+		// create our date from the timestamp
+		date := time.Unix(0, int64(ckRequest.Timestamp*1000000000)).UTC()
```
nicpottier: There's a lot of validation missing here, as to whether Timestamp or MobileNumber or RequestID is set. Probably the right way to do this is have two forms, one for receive and one for status and have validation within each. (parse and validate AFTER you look at MessageType). You can check just that one field on the request with something like:

```golang
err := r.ParseForm()
if err != nil {
    return nil, courier.WriteAndLogRequestError(ctx, w, r, c, err)
}
messageType := r.Form.Get("message_type")

// then do if blocks for the various types and use DecodeAndValidateForm() for each based on that type.
```

handlers/chikka/chikka.go
```diff
@@ -1,3 +1,148 @@
...
+	courier.RegisterHandler(newHandler())
+}
+
+// Initialize is called by the engine once everything is loaded
+func (h *handler) Initialize(s courier.Server) error {
+	h.SetServer(s)
+	return s.AddHandlerRoute(h, http.MethodPost, "receive", h.ReceiveMessage)
+}
+
+type moMsg struct {
```
nicpottier: I think I've finally settled that the right name for these is `form` at least for cases where we are getting posts or gets.

So then you'd have two forms `outgoingForm` and `incomingForm`, naming the instances `form`. Which then all works nicely with `DecodeAndValidateForm` etc..

PR #125, opened 2018-02-23
Macrokiosk handlers

norkans7:


handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,227 @@
...
+	"github.com/nyaruka/courier/gsm7"
+	"github.com/nyaruka/courier/handlers"
+	"github.com/nyaruka/courier/utils"
+	"github.com/nyaruka/gocommon/urns"
+)
+
+var (
+	sendURL                   = "https://www.etracker.cc/bulksms/send"
+	maxMsgLength              = 1600
+	configMacrokioskSenderID  = "macrokiosk_sender_id"
```
nicpottier: These should be consts.

handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,227 @@
...
+	}
+	err = s.AddHandlerRoute(h, http.MethodGet, "status", h.receiveStatus)
+	if err != nil {
+		return err
+	}
+	err = s.AddHandlerRoute(h, http.MethodGet, "receive", h.receiveMessage)
+	if err != nil {
+		return err
+	}
+	return s.AddHandlerRoute(h, http.MethodPost, "receive", h.receiveMessage)
```
nicpottier: Looking at logs, it appears these are always GET

handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,227 @@
...
+
+var statusMapping = map[string]courier.MsgStatusValue{
+	"ACCEPTED":    courier.MsgSent,
+	"DELIVERED":   courier.MsgDelivered,
+	"UNDELIVERED": courier.MsgFailed,
+	"PROCESSING":  courier.MsgWired,
+}
+
+// receiveStatus is our HTTP handler function for status updates
+func (h *handler) receiveStatus(ctx context.Context, channel courier.Channel, w http.ResponseWriter, r *http.Request) ([]courier.Event, error) {
```
nicpottier: They never seem to actually send us status messages, think we showed that as an option since we are using the external handler on RapidPro but I think we can remove it here. (and update the config on rapidpro)

handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,227 @@
...
+}
+
+type moForm struct {
+	Longcode  string `name:"longcode"`
+	Shortcode string `name:"shortcode"`
+	MSISDN    string `name:"msisdn"`
+	From      string `name:"from"`
+	Text      string `name:"text"`
+	MsgID     string `name:"msgId"`
+	Time      string `name:"time"`
```
nicpottier: Fields I see in our logs:

`GET /handlers/macrokiosk/receive/1f5c4697-16a0-4b5e-b184-07b40a76e9fb?from=60179085717&text=Pada+pendapat+saya%2c+kita+sebagai+masyarakat+mestilah+bekerjasama+dengan+pihak+sekolah+untuk+mengurangkan+kes+buli+yang+berlaku+di+sekolah.&time=2018-02-2316%3a05%3a00&shortcode=69898&telcoid=1&msgid=180223160500493421`

So only need from, text, time and msgid

handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,230 @@
...
+	}
+
+	recipient := form.Longcode
+	sender := form.MSISDN
+	if form.Shortcode != "" {
+		recipient = form.Shortcode
+		sender = form.From
+	}
+
+	if recipient == "" || sender == "" {
```
nicpottier: If we only support the fields above then you can use `validate:"required"` to simplify this stuff.

handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,230 @@
...
+	// and write it
+	err = h.Backend().WriteMsg(ctx, msg)
+	if err != nil {
+		return nil, err
+	}
+	courier.LogMsgReceived(r, msg)
+	return []courier.Event{msg}, h.writeReceiveSuccess(ctx, w, r, "-1") // MacroKiosk expects "-1" back for successful requests
+}
+
+func (h *handler) writeReceiveSuccess(ctx context.Context, w http.ResponseWriter, r *http.Request, responseText string) error {
```
nicpottier: Probably dont need to break this out if we are only calling it once.

handlers/macrokiosk/macrokiosk.go
```diff
@@ -1,5 +1,230 @@
...
+	text := gsm7.ReplaceSubstitutions(handlers.GetTextAndAttachments(msg))
+	encoding := "0"
+	if !gsm7.IsValid(text) {
+		encoding = "5"
+	}
+
+	status := h.Backend().NewMsgStatusForID(msg.Channel(), msg.ID(), courier.MsgErrored)
+	parts := handlers.SplitMsg(text, maxMsgLength)
+	for _, part := range parts {
+		payload := &mtPayload{}
```
nicpottier: If you are going to populate all of these at once, it might be cleaner to do:

```
payload := &mtPayload {
    From: senderID,
    ServID: servID,
   ...
}
```

handlers/macrokiosk/macrokiosk.go
norkans7: I checked the logs however I think the method is configurable by the user on macrokiosk side

nicpottier: Maybe then we just tell them to always use GET in our configuration?

norkans7: Ok

handlers/macrokiosk/macrokiosk.go
norkans7: the logic here is that we check the number the have shortcode or longcode and know which params to use

PR #132, opened 2018-03-01
Check urn errors

norkans7:


handlers/viber/viber_test.go
```diff
@@ -181,7 +181,7 @@ var (
 		"timestamp": 1481142112807,
 		"message_token": 4987381189870374000,
 		"sender": {
-			"id": "xy5/5y6O81+/kbWHpLhBoA==",
+			"id": "xy55y6O81kbWHpLhBoA==",
```
nicpottier: Err, are we sure these aren't valid Viber ids? Betting they come from logs.. maybe our validation is wrong in `urns`?

norkans7: right, these are right checking from the logs, I had checked the docs only

PR #137, opened 2018-03-07
FCM Channel fixes

johncordeiro:
This PR includes:

- Fixes contact urn auth when unmarshalling json message from rapidpro backend;
- Updates URN with auth for existing contacts in all the channels, not only for FCM, so please tell me if you have any concern about that. I used this approach after checking some tests (especially [this](https://github.com/nyaruka/courier/blob/master/backends/rapidpro/backend_test.go#L224) one) that the function `contactURNForURN` is called after the update of existing contacts;

### DIFF
--- a/backends/rapidpro/contact.go
+++ b/backends/rapidpro/contact.go
@@ -83,6 +83,8 @@ func contactForURN(ctx context.Context, b *backend, org OrgID, channel *DBChanne
 			tx.Rollback()
 			return nil, err
 		}
+		// update urn auth if necessary
+		contactURNForURN(tx, org, channel.ID(), contact.ID_, urn, auth)
 		return contact, tx.Commit()
 	}
 
--- a/backends/rapidpro/contact.go
+++ b/backends/rapidpro/contact.go
@@ -84,7 +84,9 @@ func contactForURN(ctx context.Context, b *backend, org OrgID, channel *DBChanne
 			return nil, err
 		}
 		// update urn auth if necessary
-		contactURNForURN(tx, org, channel.ID(), contact.ID_, urn, auth)
+		if auth != "" {
+			contactURNForURN(tx, org, channel.ID(), contact.ID_, urn, auth)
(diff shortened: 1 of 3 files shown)


backends/rapidpro/msg.go
```diff
@@ -374,7 +374,7 @@ type DBMsg struct {
 	Visibility_   MsgVisibility          `json:"visibility"      db:"visibility"`
 	HighPriority_ null.Bool              `json:"high_priority"   db:"high_priority"`
 	URN_          urns.URN               `json:"urn"`
-	URNAuth_      string                 `json:"urn_auth"`
+	URNAuth_      string                 `json:"contact_urn_auth"`
```
norkans7: the json key should be "urn_auth"

check https://github.com/nyaruka/rapidpro/blob/master/temba/channels/courier.py#L69

johncordeiro: @norkans7 Oh, you already fixed it on RapidPro, let me change back

nicpottier:
Would be good to have a unit test to demonstrate what the problem was. Was it just the key being off? Is the new call to `ContactURNForURN` needed? Would prefer to avoid that if possible since that's a new DB hit on every request otherwise.

johncordeiro:
@nicpottier [commit](https://github.com/nyaruka/courier/pull/137/commits/e4f9d39c703d020529c3100d6f44eca8b7f073ce)

backends/rapidpro/contact.go
```diff
@@ -83,6 +83,8 @@ func contactForURN(ctx context.Context, b *backend, org OrgID, channel *DBChanne
 			tx.Rollback()
 			return nil, err
 		}
+		// update urn auth if necessary
+		contactURNForURN(tx, org, channel.ID(), contact.ID_, urn, auth)
```
nicpottier: I think we only need to do this if `auth` is non-empty no?

PR #138, opened 2018-03-07
Support response_to external_id to fix chikka reply sending

norkans7:


handlers/chikka/chikka_test.go
```diff
@@ -68,13 +68,31 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
+			"request_id":    "external-id",
+			"mobile_number": "63911231234",
+			"shortcode":     "2020",
+			"request_cost":  "FREE",
+			"client_id":     "Username",
+			"secret_key":    "Password",
+			"message_id":    "10",
+		},
+		SendPrep: setSendURL},
+	{Label: "Failed Reply use Send",
```
norkans7: Not checking the Status since all requests are mocked to return 400s so it might be misleading here we are checking we retried with SEND when REPLY failed

handlers/chikka/chikka.go
```diff
@@ -129,15 +130,31 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 			form["message_type"] = []string{"REPLY"}
-			form["request_id"] = []string{msg.ResponseToID().String()}
+			form["request_id"] = []string{msg.ResponseToExternalID()}
 		}
 
 		req, _ := http.NewRequest(http.MethodPost, sendURL, strings.NewReader(form.Encode()))
 		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 		rr, err := utils.MakeHTTPRequest(req)
 
+		if rr.StatusCode == 400 {
```
nicpottier: Doh, totally missed this block when we ported this over, good catch.

PR #139, opened 2018-03-07
Allow courier to use a local S3 compatible interface

smn:
I'm finding that for certain environments, like local development, running a local S3 compatible interface like https://github.com/minio/minio is sometimes useful. 

This change allows you to use a custom S3 compatible interface for Courier's attachment handling. 
Not sure if useful outside of my use case so YMMV.


nicpottier:
+1 on this, I'm going to merge #135 first since it changes the config stuff a bit and that affects this PR. Should go in today.

smn:
@nicpottier I may be wrong but I'm getting the impression that command line parameters now no longer take precedence over the TOML config file. I'll try & verify.

nicpottier:
They should, or least that's a bug in ezconf if not.

nicpottier:
PS you can use the argument `-debug-conf` to get a dump of what ezconf is reading and the priority it applies it in.

nicpottier:
@smn https://github.com/nyaruka/courier/commit/8e3ee8cab100c66f087f53ac9dbe0e15c9c3deda should have fixed the ezconf behavior you were seeing. Was subtle, if you were overriding with a command line with the same value as the struct default, then it was ignoring it (and therefore TOML would win). Now fixed, thanks for the heads up. Now certified bug free. (TM)

smn:
Thanks! That fixed it. Let me know if you want any other changes made.

nicpottier:
Excellent, thanks!

PR #140, opened 2018-03-07
Fix chikka description in error case

norkans7:


handlers/chikka/chikka_test.go
```diff
@@ -90,7 +90,7 @@ var defaultSendTestCases = []ChannelSendTestCase{
 		Text: "Simple Message", URN: "tel:+63911231234",
 		ResponseToID:         5,
 		ResponseToExternalID: "external-id",
-		ResponseBody:         `{"status":400,"message":"BAD REQUEST","description":"Invalid/Used Request ID"}`,
+		ResponseBody:         `{"status":400,"message":"BAD REQUEST","description":"Invalid\\/Used Request ID"}`,
```
nicpottier: Does the raw text actually look like `Invalid\/Used Request ID`?

nicpottier: The python code seems to use `Invalid/Used Request ID`

norkans7: I got this by looking at the channel log

PR #141, opened 2018-03-07
check for auth changes when making sure URN is default

nicpottier:
replaces #137, @johncordeiro I think this is a bit more efficient approach, as it doesn't require an extra query on every message hit. Give it a look.


johncordeiro:
@nicpottier indeed.

nicpottier:
Cool, thanks for the bug report, debugging and PR!

PR #143, opened 2018-03-08
Add a handler to courier for Engage

smn:
Not entirely sure about this approach yet but we'll see where we end up.


smn:
Lol, this went to the wrong repo. Meant to be experimenting in my fork before I introduce anything there.

PR #146, opened 2018-04-03
Fix Zenvia Handler

johncordeiro:
This PR fixes the following bugs:

- The `sendURL` of outgoing messages was replaced to the right one according to [zenvia documentation](https://zenviasmsenus.docs.apiary.io/#reference/api-services/sending-a-single-sms/sending-a-single-sms);
- Remove the obligation of correlatedMessageSmsId, that's only sent after receiving an outgoing message before it;
- Replace the ExternalID of incoming messages to `id` instead of `correlatedMessageSmsId`;
- Remove `from` parameter of outgoing messages, which is not required and very strange when the message is received by the user: `Sender: Msg`;
- Replace callbackOptions from `1` to `FINAL` according to [zenvia documentation](https://zenviasmsenus.docs.apiary.io/#reference/api-services/sending-a-single-sms/sending-a-single-sms);


nicpottier:
Excellent, thanks. We didn't have any Zenvia channels to test with so not surprising some of that has changed.

PR #155, opened 2018-05-16
Feature Messangi Channel

johncordeiro:
This PR adds the [Messangi](http://www.messangi.com/) SMS channel. This channel was added according to direct conversations with the aggregator and following its best practices.


handlers/messangi/messangi.go
```diff
@@ -0,0 +1,132 @@
...
+	"encoding/hex"
+	"github.com/nyaruka/courier/utils"
+	"encoding/xml"
+)
+
+const (
+	configPublicKey  = "messangi_public_key"
+	configPrivateKey = "messangi_private_key"
+	configInstanceId = "messangi_instance_id"
+	configCarrierId  = "messangi_carrier_id"
```
nicpottier: no need to have these constants start with `messangi`, a channel can only be of a single type at a time.

PR #169, opened 2018-08-10
Ignore status update for incoming messsages

norkans7:


backends/rapidpro/status.go
```diff
@@ -140,6 +140,11 @@ func (b *backend) flushStatusFile(filename string, contents []byte) error {
 		return nil
 	}
 
+	// Ignore wrong status update for incoming messages
+	if err.Error() == "Incoming messages can only be PENDING or HANDLED" {
```
ericnewcomer: This should be comparing against a constant like above not doing string comparison.

PR #170, opened 2018-08-22
Added info about Redis and Rapid Pro needing to be on the same redis …

IanLawrence:
…instance and database for channel send/receive to work


nicpottier:
Even better why don't we just change the default to 15, that way people won't get confused in the same way?

PR #172, opened 2018-08-30
Log the error when failing to connect to PostgreSQL

smn:
Minor change, adds an "error" keyword to the log when failing to connect to PostgreSQL.

I got stuck trying to courier to connect to the db, turned out the COURIER_DB env var value needed a `?sslmode=disable` option. Having courier log the error would have helped me track down the issue quicker.

The error it prints now is:

```
ERRO[0000] db not reachable!     comp=backend error="pq: SSL is not enabled on the server" state=starting
```


backends/rapidpro/backend.go
```diff
@@ -377,7 +377,9 @@ func (b *backend) Start() error {
 	err = b.db.PingContext(ctx)
 	cancel()
 	if err != nil {
-		log.Error("db not reachable")
+		log.WithFields(logrus.Fields{
+			"error": err,
+		}).Error("db not reachable!")
```
nicpottier: Think the right way to do this is actually:

`log.WithError(err).Error("db not reachable")`

smn:
@nicpottier thanks, updated.

nicpottier:
:+1 thanks!

PR #177, opened 2018-09-13
clear dedupes on outgoing messages

nicpottier:
two changes that are for the better:
 - our key in our seen messages map is now channel/urn instead of channel/urn/text. the previous behavior meant that it was possible with very fast interactions to have more than one thing in each map for any channel/urn which increased the likelyhood of collisions
 - when we mark a message as sent (ie we just sent something out) we clear the seen incoming messages for a channel/urn. that should keep fast interactions with the same response from being caught

### DIFF
--- a/backends/rapidpro/msg.go
+++ b/backends/rapidpro/msg.go
@@ -308,17 +308,17 @@ func (b *backend) flushMsgFile(filename string, contents []byte) error {
 // Deduping utility methods
 //-----------------------------------------------------------------------------
 
-var luaMsgSeen = redis.NewScript(3, `-- KEYS: [Window, PrevWindow, Fingerprint]
+var luaMsgSeen = redis.NewScript(3, `-- KEYS: [Window, PrevWindow, URNFingerprint]
 	-- try to look up in window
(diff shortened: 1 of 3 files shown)


backends/rapidpro/msg.go
```diff
@@ -327,24 +327,31 @@ func checkMsgSeen(b *backend, msg *DBMsg) courier.MsgUUID {
...
-	// try to look up our UUID from either window or prev window
-	foundUUID, _ := redis.String(luaMsgSeen.Do(r, windowKey, prevWindowKey, fingerprint))
-	if foundUUID != "" {
-		return courier.NewMsgUUIDFromString(foundUUID)
+	// see if there were any messages received in the past 4 seconds
+	found, _ := redis.String(luaMsgSeen.Do(r, windowKey, prevWindowKey, urnFingerprint))
+
+	// if so, test whether the text it the same
+	if found != "" {
+		prevText := found[37:]
```
norkans7: I am confused here do we still have the text?

nicpottier: we insert the msg uuid & text in the writeMsgSeen

norkans7: I got it

PR #179, opened 2018-09-17
WhatsApp media handling

smn:
This PR adds media support for WhatsApp in courier.

This is the first time I'm writing some golang so please let me know where I'm doing things in odd ways. 

I've been fighting with the test harness a bit since it was built around the idea that API interactions would be single request & response. WhatsApp breaks that a bit since the media upload is a 2 step interaction where a media file is exchanged for a media id. I've gone the route of allowing a list of requests that the test matches against to select the expected response.


smn:
@nicpottier I would appreciate a review of this.

handlers/whatsapp/whatsapp.go
```diff
@@ -106,11 +147,42 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+			mediaURL, err = resolveMediaURL(channel, msg.Image.ID)
+		} else if msg.Type == "location" {
+			mediaURL = fmt.Sprintf("geo:%f,%f", msg.Location.Latitude, msg.Location.Longitude)
+		} else if msg.Type == "video" {
+			mediaURL, err = resolveMediaURL(channel, msg.Video.ID)
+		} else if msg.Type == "voice" {
+			mediaURL, err = resolveMediaURL(channel, msg.Voice.ID)
+		} else {
+			// we received a message type we do not support.
+			return nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r, fmt.Errorf("Unsupported message type %s", msg.Type))
```
norkans7: I think unsupported types should just be ignored and we do not want to return an error for those

handlers/whatsapp/whatsapp.go
```diff
@@ -106,11 +147,42 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+		}
+
+		// we had an error downloading media
+		if err != nil {
+			return nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r, fmt.Errorf("Error while downloading media: %s", err))
```
norkans7: Since Whatsapp sends us a list of messages and statuses we do not want to return on the first error, we need to process all of them keep record of errors and once we looped on all of them check the list of errors and return an error if we have had errors

handlers/whatsapp/whatsapp.go
```diff
@@ -88,10 +133,6 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
 
 	// first deal with any received messages
 	for _, msg := range payload.Messages {
-		// ignore it if we aren't text
-		if msg.Type != "text" {
```
norkans7: Probably we keep this if block to continue if we have an type we do not support yet

handlers/whatsapp/whatsapp.go
```diff
@@ -147,6 +219,36 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+	mediaEndpoint := url.ResolveReference(mediaPath).String()
+
+	fileURL := fmt.Sprintf("%s/%s", mediaEndpoint, mediaID)
+
+	return fileURL, nil
+}
+
+// BuildDownloadMediaRequest to download media for message attachment with Bearer token set
+func (h *handler) BuildDownloadMediaRequest(ctx context.Context, b courier.Backend, channel courier.Channel, attachmentURL string) (*http.Request, error) {
+	token := channel.StringConfigForKey(courier.ConfigAuthToken, "")
```
norkans7: What if the token is empty?

handlers/whatsapp/whatsapp.go
smn: I would recommend you keep it in though, I've seen a few cases where a message of type "contact" was received, that type's not documented by WA but would still be good to know & keep track of it for debugging purposes.

smn: but if you feel strongly about this only adding clutter then I'm happy to remove it.

handlers/whatsapp/whatsapp.go
smn: noted, will add the continue block below depending on where that bit of PR discussion goes.

handlers/whatsapp/whatsapp.go
smn: Right now it'll fail on the `Authorization`, I can add a check on non-empty though.

handlers/whatsapp/whatsapp_test.go
```diff
@@ -122,21 +263,99 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
+		ResponseBody: `{ "errors": [{"title":"Error Sending"}] }`, ResponseStatus: 200,
 		RequestBody: `{"to":"250788123123","type":"text","text":{"body":"Error"}}`,
+		Error:       "Received error from send endpoint: Error Sending",
 		SendPrep:    setSendURL},
+	{Label: "Audio Send",
+		Text:   "audio has no caption",
+		URN:    "whatsapp:250788123123",
+		Status: "W", ExternalID: "157b5e14568e8",
+		Attachments: []string{"audio/mpeg:https://foo.bar/audio.mp3"},
+		ResponseBodies: map[MockedRequest]MockedResponse{
```
smn: @norkans7 I'm not hugely happy about this attribute name, `ResponseBodies`, perhaps `Responses` is better? Any other input from you on this change. It's a fairly substantial change to your test harness and, if this is a route you're happy to take, probably can take some further refactoring to clean up the `ExternalID`, `ResponseStatus`, `ResponseBody` and `RequestBody` test case attributes.

handlers/whatsapp/whatsapp.go
smn: nevermind, seeing what you mean now, we can log it but not return.

handlers/whatsapp/whatsapp.go
smn: done

handlers/whatsapp/whatsapp.go
smn: done

handlers/whatsapp/whatsapp.go
smn: done

handlers/whatsapp/whatsapp.go
smn: done

handlers/whatsapp/whatsapp_test.go
norkans7: And yes that will require some refactor and clean up for those attributes

handlers/whatsapp/whatsapp.go
```diff
@@ -189,55 +333,150 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
-	for i, part := range parts {
-		payload := mtPayload{
-			To:   msg.URN().Path(),
-			Type: "text",
-		}
-		payload.Text.Body = part
 
-		jsonBody, err := json.Marshal(payload)
+	// TODO: figure out sending media
+	if len(msg.Attachments()) > 1 {
```
norkans7: Also I think fo the case of more that one attachments we send the first one as media with caption text and the other  we send them alone so with caption "", that is how we handle such case in Telegram, so we want to be consistent

handlers/whatsapp/whatsapp.go
```diff
@@ -147,6 +219,39 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
 	return events, courier.WriteDataResponse(ctx, w, http.StatusOK, "Events Handled", data)
 }
 
+func resolveMediaURL(channel courier.Channel, mediaID string) (string, error) {
```
norkans7: This method could alo be responsible to upload the media file and get its id to use in the message

handlers/whatsapp/whatsapp.go
smn: Cool, how do you then handle the status updates (read, delivered, etc...) for the latter ones?

handlers/whatsapp/whatsapp.go
```diff
@@ -106,11 +147,42 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+			mediaURL, err = resolveMediaURL(channel, msg.Image.ID)
+		} else if msg.Type == "location" {
+			mediaURL = fmt.Sprintf("geo:%f,%f", msg.Location.Latitude, msg.Location.Longitude)
+		} else if msg.Type == "video" {
+			mediaURL, err = resolveMediaURL(channel, msg.Video.ID)
+		} else if msg.Type == "voice" {
+			mediaURL, err = resolveMediaURL(channel, msg.Voice.ID)
+		} else {
+			// we received a message type we do not support.
+			handlers.WriteAndLogRequestError(ctx, h, channel, w, r, fmt.Errorf("Unsupported message type %s", msg.Type))
```
norkans7: I think to log only you need `courier.LogRequestError` or `courier.LogRequestIgnored` the response in this case is not needed

handlers/whatsapp/whatsapp.go
```diff
@@ -106,11 +147,42 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+		}
 
 		// create our message
-		event := h.Backend().NewIncomingMsg(channel, urn, msg.Text.Body).WithReceivedOn(date).WithExternalID(msg.ID)
+		event := h.Backend().NewIncomingMsg(channel, urn, text).WithReceivedOn(date).WithExternalID(msg.ID)
+
+		// we had an error downloading media
+		if err != nil {
+			handlers.WriteAndLogRequestError(ctx, h, channel, w, r, err)
```
norkans7: same here

handlers/whatsapp/whatsapp.go
smn: I'm not entirely sure what you're suggesting here . 

I'm happy to split out the uploading of media to WA to a different function if you feel that would be better. I do think keeping the the download stuff separate from the upload stuff is a good idea though.

handlers/whatsapp/whatsapp.go
norkans7: If a message is send as multiple parts we only save the first external ID and we'll handle the status for the first one only and ignore the others since we will not find a corresponding message with the external id we did nit save.

- So a message longer that the maxLength is send as many parts and only the first part external id is saved
- If we have many attachment we send the first with text as caption then we send the others with "" caption, also still we consider the first part only for external id and status reports

Also the upload of attachment to WA can be a different method

norkans7:
@smn I think you can add the function to upload the attachment and revert the test changes since you can test that function separately and we keep using the existing test, we can see how to make the refactor in a different PR, without  mixing both `Responses` map and the existing `ResponseBody` and other related stuff

smn:
Since the test does require a media id to be supplied, how would you suggest I mock that?

norkans7:
You mock WA service with something like https://github.com/nyaruka/courier/blob/master/handlers/telegram/telegram_test.go#L499 and that return a response with the same format as the real response

Used as https://github.com/nyaruka/courier/blob/master/handlers/telegram/telegram_test.go#L548

smn:
thanks! yeah I tried that earlier but then the setServerURL callback ended up setting the wrong server for that API call and things got quite confused. I'm happy to try that route again, I'll ping you when I get stuck on it.

handlers/whatsapp/whatsapp.go
```diff
@@ -147,6 +219,39 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
 	return events, courier.WriteDataResponse(ctx, w, http.StatusOK, "Events Handled", data)
 }
 
+func resolveMediaURL(channel courier.Channel, mediaID string) (string, error) {
+	token := channel.StringConfigForKey(courier.ConfigAuthToken, "")
+	if token == "" {
+		return "", fmt.Errorf("Missing token for WA channel")
```
nicpottier: minor stylistic issue error messages are always lowercase in golang.

smn:
@nicpottier & @norkans7 thanks for the review. I've confirmed that the current implementation does work. I'll attend to your comments coming Tuesday, I'm taking a few days leave.

handlers/whatsapp/whatsapp.go
```diff
@@ -189,55 +333,150 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+	} else if len(msg.Attachments()) == 1 {
+
+		attachment := msg.Attachments()[0]
+		parts := strings.SplitN(attachment, ":", 2)
+		mimeType := parts[0]
+		s3url := parts[1]
+
+		// retrieve the media to be sent from S3
+		req, _ := http.NewRequest(http.MethodGet, s3url, nil)
+		s3rr, err := utils.MakeHTTPRequest(req)
```
norkans7: I had just mentioned that on Slack, here we know we can allow all content types so we do something like 
https://github.com/nyaruka/courier/blob/master/backends/rapidpro/msg.go#L215

And keep the `isText` flag in `newRRFromResponse`

norkans7: I would also make a function that handle one attachment work here, download from s3 and upload to WA that will return the mediaId that can be used in the the payload below

smn:
@nicpottier I'm back from leave tomorrow and I'm happy to attend to the outstanding requested changes.

nicpottier:
Ah ok, no worries or rush just wasn't sure if your other commits had addressed everything and you wanted me to take another look. Just @ me when you are ready and we'll get this merged in!

smn:
@norkans7 & @nicpottier I think I've addressed most of the issues. Let me know if you're happy to have this merged in.

utils/http.go
```diff
@@ -135,13 +135,11 @@ func newRRFromResponse(method string, requestTrace string, r *http.Response) (*R
 	}
 	rr.Response = string(response)
 
-	if isText {
-		bodyBytes, err := ioutil.ReadAll(r.Body)
-		if err != nil {
-			return &rr, err
-		}
-		rr.Body = bodyBytes
+	bodyBytes, err := ioutil.ReadAll(r.Body)
```
norkans7: Can we revert this in favor of the described approach on https://github.com/nyaruka/courier/pull/179#discussion_r218861360?

I think we can keep this place expecting text responses and only allow other content types in specific places like the upload to WA 

@nicpottier thoughts?

nicpottier: What's your worry here @norkans7? Seems having this be generic is probably just fine since callers are still going to try to parse the body one way or the other no?

nicpottier: (that said, doesn't seem we need the `isText` variable anymore.

norkans7: This function being used with the code facing the aggregators for text I think it is more safe with this check

nicpottier: In what way? It wasn't like we were returning an error previously when we got binary data, we just had an empty body, which seems just as likely to blow up as a binary body. All the handlers that are actually inspecting the body are parsing it in a way that should be just fine from that respect. (in that they will fail in the same way with an empty body as a binary body, such as parsing json for example) Do you have any specific example that show what you are worried about?

norkans7: I think we could be wasting time to download a 'big file' while we will not use it
really I though that was the reason to have it around

nicpottier: It is, but that's more of a vestige from goflow which this originally came from I think. We are hitting known endpoints with courier and we shouldn't be getting gigungous replies so we should be fine in that respect. I think not having to repeat all our junk around headers, reading body etc, is worth it from a reuse perspective, so I think reading the body here even in the binary case is just fine.

smn:
@nicpottier & @norkans7 happy with me leaving it as is? The last change I pushed was one to make sure the User-Agent is included in the HTTP requests being made to the WhatsApp API endpoints.

handlers/whatsapp/whatsapp.go
```diff
@@ -63,6 +64,50 @@ type eventPayload struct {
 		Text      struct {
 			Body string `json:"body"`
 		} `json:"text"`
+		Audio struct {
```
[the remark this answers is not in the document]
smn: oh wait, I forgot to make these pointers, I only did that for the outbound message path.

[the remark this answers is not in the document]
smn: done

handlers/whatsapp/whatsapp.go
```diff
@@ -63,6 +64,50 @@ type eventPayload struct {
 		Text      struct {
 			Body string `json:"body"`
 		} `json:"text"`
+		Audio *struct {
+			File     string `json:"file"`
+			ID       string `json:"id"`
+			Link     string `json:"link"`
+			MimeType string `json:"mime_type"`
+			Sha256   string `json:"sha256"`
```
nicpottier: So now that the outer is optional as a pointer, you can add some validation on the inner parts. (ie `validate:"required" where appropriate) That'll keep surprising things from happening downstream.

smn: good point, I'll add.

smn:
No, that media handling is now a lot more sane and wrapped in an HTTP API. Nothing else is needed there now.

handlers/whatsapp/whatsapp.go
smn: done, I think.

PR #186, opened 2018-10-17
🏆Support checking configured response content for EX channels

norkans7:


handlers/external/external.go
```diff
@@ -262,7 +264,9 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 			return status, nil
 		}
 
-		status.SetStatus(courier.MsgWired)
+		if responseContent == "" || strings.Contains(string(rr.Body), responseContent) {
+			status.SetStatus(courier.MsgWired)
+		}
```
nicpottier: Should probably add some context here in the status log that we failed because we didn't find the expected content. Otherwise we will probably be scratching our heads wondering why these fail.

norkans7: Updated

PR #190, opened 2018-10-29
🎈 Better channel logs support for WA channels

norkans7:
First step so we can have proper logs that will better explain why we do not get the expected ID,
Then we can fix that

### DIFF
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -337,12 +337,12 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 
 	if len(msg.Attachments()) > 0 {
 		for attachmentCount, attachment := range msg.Attachments() {
+			var log *courier.ChannelLog
 
 			mimeType, s3url := handlers.SplitAttachment(attachment)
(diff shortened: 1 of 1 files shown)


handlers/whatsapp/whatsapp.go
```diff
@@ -337,12 +337,12 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 
 			mimeType, s3url := handlers.SplitAttachment(attachment)
-			mediaID, err := uploadMediaToWhatsApp(mediaURL, token, mimeType, s3url)
+			mediaID, log, err := uploadMediaToWhatsApp(msg, mediaURL, token, mimeType, s3url)
 			if err != nil {
-				duration := time.Now().Sub(start)
-				log := courier.NewChannelLogFromError("Unable to upload media to WhatsApp server", msg.Channel(), msg.ID(), duration, err)
+				log.WithError("Unable to upload media to WhatsApp server", err)
 				status.AddLog(log)
 				return status, err
```
nicpottier: can `status.AddLog(log)` be outside all these conditions?

nicpottier:
Couple small tweaks but looks good.

PR #193, opened 2018-11-14
🖼 Add support for media attachment for Twitter channels

norkans7:


handlers/test.go
```diff
@@ -266,6 +268,10 @@ func RunChannelSendTestCases(t *testing.T, channel courier.Channel, handler cour
 				require.Equal(testCase.RequestBody, strings.Trim(string(value), "\n"))
 			}
 
+			if (len(testCase.Responses)) != 0 {
+				require.Equal(mockRRCount, len(testCase.Responses))
```
norkans7: We should make sure all the request we expected are all made

PR #194, opened 2018-11-16
PlayMobile Channel (Uzbekistan)

johncordeiro:
This PR adds the SMS Aggregator PlayMobile (http://playmobile.uz) from Uzbekistan.


handlers/playmobile/playmobile.go
```diff
@@ -0,0 +1,185 @@
...
+	"strings"
+	"encoding/json"
+
+	"github.com/nyaruka/courier"
+	"github.com/nyaruka/courier/handlers"
+	"github.com/nyaruka/courier/utils"
+)
+
+const (
+	configShortCode  = "shortcode"
```
norkans7: we normally save the shortcode in the address field, why is it in config here?

handlers/playmobile/playmobile.go
```diff
@@ -0,0 +1,185 @@
...
+	configUsername = "username"
+	configPassword = "password"
+)
+
+var (
+	maxMsgLength = 160
+	sendURL      = "http://91.204.239.42:8083/broker-api/send"
```
norkans7: No domain name?

edudouglas: Added "base_url" in channel config to save URL.

edudouglas:
> It seems this can be achieved using the existing EX channel by configuring properly the configuration.

PlayMobile works with JSON to incoming (method SendMsg) and XML to output (method receiveMessage).

norkans7:
If think all that is supported by the EX channel type and we do not need a
new channel type for this. Check how to use from_xpath and text_path

edudouglas:
@norkans7,  On Courier's EX channel, there's "from_xpath" and "text_xpath". However, the same channel on RapidPro/RapidPro there's no place to save those configurations. How could be set that the incoming is an XML and outcoming is JSON? Could you please give me more details about it?

norkans7:
For MT the JSON can be set in body which exists already.
For MO the XML from_xpath and text_xpath, there is no web UI yet, but those can be set on the shell for now and make that work, we have not yet added the UI. So may be that could be the contribution to make instead of adding this as a new channel type

johncordeiro:
hey @norkans7, once that's just a hidden feature today, I would like to request you guys to accept this PR for now. I totally agree with you about the improvement on External API to avoid channel creation and we'll start contributing on that, by the way, can you guys create an issue explaining how you are thinking to implement that?

Please accept this PR for now.

norkans7:
@johncordeiro I can clarify that the functionality is already supported 

https://github.com/nyaruka/courier/blob/master/handlers/external/external.go#L132
we use https://github.com/antchfx/xmlquery to get the value of the configured element from the XML

I just said that there is no Web view for that yet, it is only modified on the shell for now

I opened an issue to discuss how to make that used on the web https://github.com/rapidpro/rapidpro/issues/823

For this PR so far I think it is not needed to have a new channel type since we can make the channel work with EX

norkans7:
@johncordeiro @edudouglas any luck using the EX channel type?
IMO I think we need to use the existing types if those can work and see how to improve them to make the UX better and easy to use. However we'll leave this PR open so others can review it too.

handlers/playmobile/playmobile.go
```diff
@@ -0,0 +1,191 @@
...
+		message.Recipient = strings.TrimLeft(msg.URN().Path(), "+")
+		message.MessageID = msg.ID().String()
+		message.SMS.Originator = shortCode
+		message.SMS.Content.Text = part
+
+		payload.Messages = append(payload.Messages, message)
+
+		jsonBody, err := json.Marshal(payload)
+		if err != nil {
+			return status, err
```
nicpottier: This shouldn't be returning an error, rather it should be setting the error on the status then returning that. (we should only return errors in send methods for things that are internal errors (ie, DB down, misconfiguration etc))

edudouglas: Please, @nicpottier review the changes, thank you.

nicpottier:
@norkans7  @johncordeiro  @edudouglas So on the subject of using EX channels vs creating new ones, I actually think we need to keep in mind that the goal of channels is to make integrations easy for end users. It is unreasonable to expect them to understand all the parameters around configuring external channels, so creating custom channels isn't at all a bad thing, it makes people picking up RapidPro for their project easier. That said, we should be striving to reuse code as much as possible, so if we can make adding a new channel type be essentially adding a set of configurations, then that's the ideal, easy to use for the end user, but without much new code.

So really we should be trying to create new base handlers that work with XML for example so that new handlers in courier are just a few lines of code, similar to say how say receiving works for ArabiaCell by using the `NewTelReceiveHandler` function.

In any case, that doesn't need to apply to this PR, but did want to clarify the vision / reasoning around adding new channel types. While you can use external channels for many integrations, and we do for ones where we don't expect much reuse, the gold standard is indeed to create a new channel type and we should encourage that.

handlers/playmobile/playmobile.go
nicpottier: @edudouglas I am so sorry but apparently having just come back from vacation I wasn't thinking clearly here, I though this was unmarshalling a response rather than building a request. Having `Marshal` fail does indeed fall into the category of things that are internal errors (it should never happen unless we run out of ram or something) so just returning the error (with a nil status) is indeed a valid thing to do here. Can you revert your change and modify this to just return `nil, err`?

Again, apologies for the mistake on my side.

norkans7:
Okay, I understand

handlers/playmobile/playmobile.go
edudouglas: No problems @nicpottier. I reverted the code.

PR #196, opened 2018-12-03
🖇 Fix twitter attachments

norkans7:


handlers/twitter/twitter.go
```diff
@@ -400,30 +421,67 @@ func uploadMediaToTwitter(msg courier.Msg, mediaUrl string, attachmentMimeType s
 	twReq.Header.Set("Accept", "application/json")
 	twReq.Header.Set("User-Agent", utils.HTTPUserAgent)
 	twrr, err := utils.MakeHTTPRequestWithClient(twReq, client)
+	log = courier.NewChannelLogFromRR("Media Upload INIT", msg.Channel(), msg.ID(), twrr)
 	if err != nil {
-		return "", courier.NewChannelLogFromRR("Media Upload Error INIT", msg.Channel(), msg.ID(), twrr), err
+		log.WithError("Media Upload INIT Error", err)
```
nicpottier: This method is a bit of a mess with all the different logs created etc.. Should be refactored to just return an error and the caller should be responsible for wrapping that in a channel log. We should never have this much boilerplate, that's a sign that something needs to be rethought.

PR #198, opened 2018-12-03
Add optional charset channel config for EX channels

norkans7:


nicpottier:
Not sure I understand this.. why isn't this just part of the URL or the post body?

norkans7:
That is what the aggregator expect there

norkans7:
Replaced by #199

PR #207, opened 2019-01-23
🌍Adjust BL channels to use a single URL

norkans7:


handlers/bongolive/bongolive.go
```diff
@@ -35,41 +36,73 @@ func init() {
...
+	7:  courier.MsgErrored,
+	8:  courier.MsgSent,
+	9:  courier.MsgErrored,
+	10: courier.MsgErrored,
+	11: courier.MsgErrored,
+}
+
 type moForm struct {
-	ID      string `name:"ID"`
+	ID      string `name:"ID" validate:"required"`
```
nicpottier: minor but I like spacing these out so all the `validate` tags are aligned.

PR #210, opened 2019-02-05
Properly handle long attachment description for Viber

norkans7:


handlers/viber/viber_test.go
```diff
@@ -95,6 +95,17 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
+	{Label: "Long Description with Attachment",
+		Text: "Text description is longer that 10 characters",
+		URN:  "viber:xy5/5y6O81+/kbWHpLhBoA==", Attachments: []string{"image/jpeg:https://localhost/image.jpg"},
+		Status: "W", ResponseStatus: 200,
+		ResponseBody: `{"status":0,"status_message":"ok","message_token":4987381194038857789}`,
+		Headers: map[string]string{
+			"Content-Type": "application/json",
+			"Accept":       "application/json",
+		},
+		RequestBody: `{"auth_token":"Token","receiver":"xy5/5y6O81+/kbWHpLhBoA==","text":"description is longer that 10 characters","type":"text","tracking_data":"10"}`,
```
nicpottier: Is this doing the right thing? Wouldn't we expect the last message to just be the end of the message as opposed to being all of it except for `Text`?

norkans7: I mocked descripionMaxLength to 10 which is in the middle of a word. So this is right, what number will make that more clear?

nicpottier: Isn't the text `Text description is longer that 10 characters`? Wouldn't character 10 be somewhere in the middle of `description`?

norkans7: correct and our split is smart and find the previous `space` https://github.com/nyaruka/courier/blob/master/handlers/utils.go#L102

nicpottier: Oh oh oh, of course. Ok, nice work then!

PR #214, opened 2019-02-15
send fb attachments first instead of last, add quick replies to last message instead of first

nicpottier:
Addresses: https://app.intercom.io/a/apps/l6e7tr01/inbox/inbox/155915/conversations/20846125919

@norkans7 do you know why we did it the other way around? seems odd.


norkans7:
Looks good, I guess it was because we used to support one attachment only
and the text was a caption

PR #219, opened 2019-02-20
stop looking / writing is_test

nicpottier:
We'll have to bump our major version number for this as it will break others.


nicpottier:
(not going to merge until we push resist)

PR #222, opened 2019-03-07
👐Send configured welcome message on conversation started for Viber

norkans7:


handlers/viber/viber.go
```diff
@@ -96,7 +105,42 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+			return nil, handlers.WriteAndLogRequestIgnored(ctx, h, channel, w, r, "ignored conversation start")
+		}
+		// build the channel event
+		channelEvent := h.Backend().NewChannelEvent(channel, courier.WelcomeMessage, urn).WithContactName(ContactName)
+
+		err = h.Backend().WriteChannelEvent(ctx, channelEvent)
+		if err != nil {
+			return nil, err
+		}
+		payload := welcomeMessagePayload{
```
nicpottier: would move this and below into `writeWelcomeMessage Response` (don't need Viber, already in Viber package)

nicpottier:
We'll also need to tweak mailroom to receive these events. Important bit here is that I think this should be considered a new contact, so mailroom has to act accordingly (process dynamic groups, etc)

handlers/viber/viber.go
```diff
@@ -22,11 +22,12 @@ import (
...
-	sendURL              = "https://chatapi.viber.com/pa/send_message"
-	maxMsgLength         = 7000
-	quickReplyTextSize   = 36
-	descriptionMaxLength = 120
+	viberSignatureHeader      = "X-Viber-Content-Signature"
+	sendURL                   = "https://chatapi.viber.com/pa/send_message"
+	maxMsgLength              = 7000
+	quickReplyTextSize        = 36
+	descriptionMaxLength      = 120
+	configViberWelcomeMessage = "welcome_message"
```
nicpottier: things we don't change in tests should be `const`

handlers/viber/viber.go
```diff
@@ -96,7 +105,28 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
...
+			return nil, handlers.WriteAndLogRequestIgnored(ctx, h, channel, w, r, "ignored conversation start")
+		}
+
+		viberID := payload.User.ID
+		ContactName := payload.User.Name
+
+		// build the URN
+		urn, err := urns.NewURNFromParts(urns.ViberScheme, viberID, "", "")
+		if err != nil {
+			return nil, handlers.WriteAndLogRequestIgnored(ctx, h, channel, w, r, "ignored conversation start")
```
nicpottier: Should log this as an error as well, shouldn't be a normal case where this wouldn't be valid.

PR #225, opened 2019-03-20
Don't save duplicate messages sent by the WA business API

erikh360:
We've had a few cases where the WA business API sends us the same message more than once.

The flows in Rapidpro responds to both messages which is not a desirable outcome for us.

This PR will check if the message was already saved and ignore it if it was.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -50,6 +50,9 @@ type Backend interface {
 	// WriteMsgStatus writes the passed in status update to our backend
 	WriteMsgStatus(context.Context, MsgStatus) error
 
+	// CheckMsgExternalIDExistsInDB checks if a message exists with the given external ID
+	CheckMsgExternalIDExistsInDB(string) (int, error)
+
 	// NewChannelEvent creates a new channel event for the given channel and event type
 	NewChannelEvent(Channel, ChannelEventType, urns.URN) ChannelEvent
 
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -269,6 +269,11 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus) 
 	return nil
 }
 
+// CheckMsgExternalIDExistsInDB checks if a message exists with the given external ID
+func (b *backend) CheckMsgExternalIDExistsInDB(externalID string) (int, error) {
+	return checkMsgExternalIDExistsInDB(b, externalID)
+}
+
 // NewChannelEvent creates a new channel event with the passed in parameters
 func (b *backend) NewChannelEvent(channel courier.Channel, eventType courier.ChannelEventType, urn urns.URN) courier.ChannelEvent {
 	return newChannelEvent(channel, eventType, urn)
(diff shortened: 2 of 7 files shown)


erikh360:
@nicpottier @norkans7 mind having a look and advising? Thanks

nicpottier:
Have you talked to WhatsApp about this? How far apart are the messages?

There's already some deduping taking place in NewIncomingMsg (see `checkMsgSeen`). Doing a DB query on every incoming message probably isn't something we would merge into mainline, it won't do very well under high load, but depending on how far apart these are perhaps that method can be tweaked.

erikh360:
Thanks for the response, I'll get back to you on the first question.

The messages were just over 2 minutes apart:
`16-03-19 15:37:34:998`
`16-03-19 15:35:10:444`

norkans7:
Are those submitted in the same request? If so we could ignore seen IDs in payload.Messages

erikh360:
I don't think they came from the same request.

smn:
Yeah these arrived in separate requests but with the same message id.

@nicpottier I've not chatted to WA about this but I'll file a support request. A year ago they didn't make any commitments with regard to exactly-once or at-least-ones delivery guarantees but I'll ask and report back.

smn:
I've traced this down in the coreapp's logs to two separate webhooks being dispatched by the coreapp for the same message. This wasn't a retry due to a failed first delivery. I've submitted the relevant logs to WA and have asked them what their delivery guarantees are, if any.

nicpottier:
These have the same whatsapp ID though right? Maybe we can add a ~5 minute window using external ID and dedupe based on that?

smn:
@nicpottier yeah, they have the same external ID. A dedupe based on a time window makes sense if you're happy with that.

nicpottier:
Ya, I think an implementation similar to `checkMsgSeen`  as an interface on Backend probably makes the most sense. I don't think every channel type should be deduping by external ID because I think the rules around that are probably different per type (IE, I wouldn't be surprised if some providers use same external ID for parts of a multipart). So seems like something that should be an interface on Backend `CheckExternalIDSeen(id string) (uuid, err)` that does something similar to checkMsgSeen in that it checks a few expiring maps in Redis and returns the UUID if found. After a write of the msg to the db, WhatsApp handler then calls `MarkExternalIDSeen(id string) err` to stuff it in the map.

smn:
Heard back from WA on this:

> ... the guarantee would be for it to be delivered at-least-once, and even our consumer apps client perform deduplication on the client based on the message id. There are many reasons why our server could see a request sent to your client as failed or timed out, while it was in fact "successful enough" from your server to be processed; and in this case it will retry sending it.

We're seeing re-deliveries of up until 19 hours later. I'm assuming those aren't so problematic in RapidPro as the flow would generally have timed out before then.

Based on our stats, having a 5 minute window within we would duplicate would catch ~27% of the redeliveries. We may want to make that window configurable depending on the use case. Some of our data science teams' work is concerned about having double deliveries invalidate some of their surveys and they would rather sacrifice some performance than the research results.

norkans7:
The solution @nicpottier suggested is to follow the approach of `checkMsgSeen` and use Redis instead of checking the DB, then configure a 24 hours window for the external IDs

nicpottier:
Yes, that's right. Fine deduping for the WA case but shouldn't hit the DB to do it.

erikh360:
Thanks for the feedback everyone, I'll make the changes.

erikh360:
I've created a new PR for the suggested changes: https://github.com/nyaruka/courier/pull/227

PR #227, opened 2019-03-28
Deduplicate WA messages on external ID

erikh360:
We've had a few cases where the WA business API sends us the same message more than once.

The flows in Rapidpro responds to both messages which is not a desirable outcome for us.

This PR will deduplicate messages based on the external ID on the WA channel.

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -296,6 +296,16 @@ func (b *backend) WriteChannelLogs(ctx context.Context, logs []*courier.ChannelL
 	return nil
 }
 
+// Check if external ID has been seen in a period
+func (b *backend) CheckExternalIDSeen(externalID string) bool {
+	return checkExternalIDSeen(b, externalID)
+}
+
+// Mark a external ID as seen for a period
+func (b *backend) WriteExternalIDSeen(externalID string) {
+	writeExternalIDSeen(b, externalID)
+}
+
 // Health returns the health of this backend as a string, returning "" if all is well
 func (b *backend) Health() string {
 	// test redis
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -297,13 +297,20 @@ func (b *backend) WriteChannelLogs(ctx context.Context, logs []*courier.ChannelL
 }
 
 // Check if external ID has been seen in a period
-func (b *backend) CheckExternalIDSeen(externalID string) bool {
-	return checkExternalIDSeen(b, externalID)
+func (b *backend) CheckExternalIDSeen(msg courier.Msg) courier.Msg {
+	var prevUUID = checkExternalIDSeen(b, msg)
+	m := msg.(*DBMsg)
+	if prevUUID != courier.NilMsgUUID {
+		// if so, use its UUID and that we've been written
+		m.UUID_ = prevUUID
+		m.alreadyWritten = true
+	}
+	return m
 }
 
 // Mark a external ID as seen for a period
-func (b *backend) WriteExternalIDSeen(externalID string) {
-	writeExternalIDSeen(b, externalID)
+func (b *backend) WriteExternalIDSeen(msg courier.Msg) {
+	writeExternalIDSeen(b, msg)
 }
 
 // Health returns the health of this backend as a string, returning "" if all is well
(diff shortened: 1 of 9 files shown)


backends/rapidpro/msg.go
```diff
@@ -401,6 +401,52 @@ func clearMsgSeen(rc redis.Conn, msg *DBMsg) {
...
+`)
+
+func writeExternalIDSeen(b *backend, externalID string) {
+	r := b.redisPool.Get()
+	defer r.Close()
+
+	now := time.Now().In(time.UTC)
+	windowKey := fmt.Sprintf("seen:externalid:%s", now.Format("2006-01-02"))
+
+	luaWriteExternalIDSeen.Do(r, windowKey, externalID, true)
```
norkans7: I think here it is important we use 
* the `msg.urnFingerprint()` then append the external ID to it 
* the window key should be more precise than the date alone
* we need to be able to not ignore the request if the message text is different even if for any reason the external ID is the same

handlers/whatsapp/whatsapp.go
```diff
@@ -134,6 +134,13 @@ func (h *handler) receiveEvent(ctx context.Context, channel courier.Channel, w h
 	// first deal with any received messages
 	for _, msg := range payload.Messages {
 
+		if msg.ID != "" {
```
norkans7: The best place to check external ID is seen is https://github.com/nyaruka/courier/pull/227/files#diff-3b05f66223be3994d25cf04fda0cfca0L175 and use `alreadyWritten` field to control the if we need to write to the DB

erikh360: Thanks for the review. 

Do you mean inside the `NewIncomingMsg` function? I kept it outside to only do it in the WA handler.

norkans7: Not inside rather after the call on that line, I agree we need this in the WA handle only

backends/rapidpro/msg.go
nicpottier: Honestly I think what's here is sufficient. This is only used by channels that are asserting that external ids are canonical so they can make the call that even if the msg body is different then it is fine.

Do agree however that they should probably be keyspaced to the channel, so the set you are adding to should use the channel UUID. But otherwise I think this is ok.

nicpottier: Also we should be putting the UUID of the message in the map so that we keep the contract of returning the right UUID in our response. (same as checkMsgSeen)

handlers/whatsapp/whatsapp.go
nicpottier: Ya, main thing here I think is that we do want to echo the message back from a courier contract perspective just with the UUID of the previous message.

backends/rapidpro/msg.go
erikh360: @norkans7 for `the window key should be more precise than the date alone`. I'm not sure how to go about this for a 24h window. 

If I include the hour in the key then I'll end up with more than two buckets, any suggestion on how you would approach this?

norkans7: @erikh360 never mind on the window, the date is enough

erikh360: @norkans7 Thanks, will you re-review my latest please?

backends/rapidpro/backend.go
```diff
@@ -296,6 +296,23 @@ func (b *backend) WriteChannelLogs(ctx context.Context, logs []*courier.ChannelL
 	return nil
 }
 
+// Check if external ID has been seen in a period
+func (b *backend) CheckExternalIDSeen(msg courier.Msg) courier.Msg {
+	var prevUUID = checkExternalIDSeen(b, msg)
+	m := msg.(*DBMsg)
```
norkans7: is this conversion needed?

erikh360: Yes, as far as I understand, `alreadyWritten` is not defined on `courier.Msg`, only on `DBMsg`.

norkans7: Okay, actually that is fine since `DBMsg` implements the interface we are returning `courier.Msg`

erikh360:
@nicpottier you happy with my latest changes?

norkans7:
one last thing, can you increase the coverage?

erikh360:
@norkans7 looks like the coverage only dropped on `test.go`, not sure if that should be included in the check.

nathanbegbie:
Could we add `**/*test.go` to `codecov.yml` ?

norkans7:
Sorry for the late reply, no need to increase the coverage if it is in the tests file
I think to ignore that file you can use `test.go`

@nicpottier please review as well

nicpottier:
Looks good, thanks for this and working through it @erikh360

PR #231, opened 2019-04-11
Add config for maintenance mode, Skip write to DB and flushing spool …

norkans7:
…and write to spool only


ericnewcomer:
Why no redis? Seems duplicate checking on spooling is still valuable no?

backends/rapidpro/msg.go
```diff
@@ -72,14 +73,19 @@ func writeMsg(ctx context.Context, b *backend, msg courier.Msg) error {
...
 
 	// try to write it our db
-	err := writeMsgToDB(ctx, b, m)
+	if !b.config.Maintenance {
+		err = writeMsgToDB(ctx, b, m)
+	}
 
-	// fail? spool for later
+	// fail? log
 	if err != nil {
```
ericnewcomer: what err is this checking?

norkans7:
I thought that is not useful since the UUID are nil in that case also we use the current timestamp for spool files so I thought the seen should written only when not in maintenance mode

norkans7:
But I guess we can make redis available for the external ID seen only

PR #242, opened 2019-06-05
Add SignalWire handler

nicpottier:
https://signalwire.com/

Mostly the same API as Twilio except they don't sign.

Renamed `twilio` handler to `twiml` since that's more accurate now.

Also removed global 'ignore_delivery_reports' setting which was only used by Resist. That's now a channel config. 

@chris-erickson, I set `ignore_dlrs` to `true` on the config for the main ResistBot Twilio channel so this doesn't change anything for you. (that global config option was only for resist)

### DIFF
--- a/backends/rapidpro/channel.go
+++ b/backends/rapidpro/channel.go
@@ -241,6 +241,16 @@ func (c *DBChannel) StringConfigForKey(key string, defaultValue string) string {
 	return str
 }
 
+// BoolConfigForKey returns the config value for the passed in key, or defaultValue if it isn't found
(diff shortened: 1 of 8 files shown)


chris-erickson:
What do you mean ` I set ignore_dlrs to true on the config for the main ResistBot Twilio channel`?  Thanks for no change though 😅

nicpottier:
I mean I already changed the config on resist for you.

tybritten:
@nicpottier can you expand on that? I'm assuming you mean you added `ignore_dlrs=true` somewhere in the resist config..where?

nicpottier:
Every channel has a JSON config field, I set it on the resist shortcode channel.

PR #253, opened 2019-08-14
Add adress and domain to the default configuration

pangoSE:
Without these the courier does not start listening, it seems.

### DIFF
--- a/courier.toml
+++ b/courier.toml
@@ -9,6 +9,12 @@
 #
 #-------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


nicpottier:
Appreciate the PR but courier does listen by default, it is listening on localhost, which is different than all network interfaces (`0.0.0.0`) and the correct thing if you are putting proxies in front of courier. Current config is correct, you can always do `--help` to see all possible options, the `.toml` is not mean to be expansive. (we recommend using environment variables in any case)

PR #254, opened 2019-08-16
Check whether outgoing message is in a loop before sending

nicpottier:
This basically means that any time a contact gets the same flow message more than 20 times in a 5 minute window we will fail it. I *think* that's reasonable.

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -208,6 +208,9 @@ var luaMsgLoop = redis.NewScript(3, `-- KEYS: [key, contact_id, text]
 	-- write our new record witn updated count
 	redis.call("hset", key, contact_id, record)
 
+	-- sets its expiration
+	redis.call("expire", key, 300)
+
 	return count
 `)
 
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -164,7 +164,7 @@ func (b *backend) PopNextOutgoingMsg(ctx context.Context) (courier.Msg, error) {
 }
 
 var luaSent = redis.NewScript(3,
-	`-- KEYS: [TodayKey, YesterdayKey, MsgId]
+	`-- KEYS: [TodayKey, YesterdayKey, MsgID]
      local found = redis.call("sismember", KEYS[1], KEYS[3])
      if found == 1 then
 	   return 1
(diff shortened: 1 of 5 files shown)


backends/rapidpro/backend.go
```diff
@@ -183,6 +183,63 @@ func (b *backend) WasMsgSent(ctx context.Context, msg courier.Msg) (bool, error)
...
+			count = tonumber(record_count) + 1
+		else
+			count = 1
+		end		
+	end
+
+	-- create our new record with our updated count
+	record = string.format("%02d:%s", count, text)
+
+	-- write our new record witn updated count
```
rowanseymour: typo

backends/rapidpro/backend.go
```diff
@@ -183,6 +183,63 @@ func (b *backend) WasMsgSent(ctx context.Context, msg courier.Msg) (bool, error)
...
+	defer rc.Close()
+
+	keyTime := time.Now().UTC().Round(time.Minute * 5)
+	key := fmt.Sprintf(sentSetName, fmt.Sprintf("loop_msgs:%s", keyTime.Format("2006-01-02-15:04")))
+	count, err := redis.Int(luaMsgLoop.Do(rc, key, m.ContactID_, m.Text_))
+	if err != nil {
+		return false, errors.Wrapf(err, "error while checking for msg loop")
+	}
+
+	if count >= 20 {
```
rowanseymour: const ?

sender.go
```diff
@@ -177,21 +177,34 @@ func (w *Sender) sendMessage(msg Msg) {
...
-		msgLog.WithError(err).Warning("error looking up msg was sent")
+		log.WithError(err).Warning("error looking up msg was sent")
+	}
+
+	// is this msg in a loop?
+	loop, err := backend.IsMsgLoop(sendCTX, msg)
+
+	// failing on loop lookup isn't permanent, but log
+	if err != nil {
+		log.WithError(err).Warning("error looking up msg loop")
```
rowanseymour: if this is just a warning, any chance it stops working and we never notice?

nicpottier: I *think* warnings still show up in Sentry, but actually I do prefer these being real errors from a sentry perspective, will change.

backends/rapidpro/backend.go
nicpottier: If something is only used once I feel like naming it into a const is just hiding it, but can change.

rowanseymour: that's fair - was just thinking from a "advertising that something is configurable" pov

PR #263, opened 2019-10-29
Send WA media by web URL

norkans7:


norkans7:
This was added in 2.23.5(available until Dec 15, 2019) and 2.23.4 was available until Oct 26, 2019

norkans7:
current version is 2.25.3

nicpottier:
Oh nice, in that case this seems safe to make the default unless @smn speaks up.

smn:
Yeah this should work fine on pretty much all installations. Where it could fail is where an outdated version is still running as it'll only complain about the build's expiry date on a restart. If it's kept running then it won't support the "link" parameter yet. Given the 26th of October was only a few days ago, this could still impact a couple of installations.

Are there plans on to support the [provider](https://developers.facebook.com/docs/whatsapp/api/settings/media-providers) field as well to allow for access control om media backends? 

[Turn](https://turn.io) doesn't support that yet, mostly because no-one's asked for that feature and quite frankly seems fairly over-engineered.

Issue #262, opened 2019-10-29
Error updating status on incoming message


nicpottier:
See: https://sentry.io/organizations/nyaruka/issues/1302032063/?project=205713&referrer=alert_email

Pretty sure this is us receiving a `sent` status for an incoming message, in this case that message was one with an attachment. (msg id 68258595 in textit)

This isn't causing any problems except the sentry error, but we should deal with it:

1. Doesn't look like we are writing a channel log for the event that is coming in, that's a bug
2. We should change our SQL for updating status based in external ID to filter by direction to avoid the integrity error

norkans7:
I looked into this and I can see the 2nd issue but on the first one it seems we never write any logs for spooled objects.

Do we need to add a similar spool approach for the channel logs too?

norkans7:
with #265 I think 1 is fixed too since it was caused by the spooled statuses from https://github.com/nyaruka/courier/blob/master/batch/batch.go#L144 the same integrity error thrown by the DB (Incoming messages can only be PENDING or HANDLED) for the bulk update

PR #265, opened 2019-10-30
Fix error trying to update incoming message status

norkans7:


norkans7:
Fixes https://github.com/nyaruka/courier/issues/262

PR #266, opened 2019-11-25
🏷️ Facebook message type tags

rowanseymour:
See https://developers.facebook.com/docs/messenger-platform/send-messages/#messaging_types and https://developers.facebook.com/docs/messenger-platform/send-messages/message-tags/#current_supported_tags


handlers/facebook/facebook.go
```diff
@@ -37,6 +37,14 @@ var (
 		369239343222814: "👍", // medium
 		369239383222810: "👍", // big
 	}
+
+	tagByTopic = map[string]string{
+		"":         "ACCOUNT_UPDATE",
```
rowanseymour: need a default for a msg with no topic

rowanseymour: or do we keep sending with `NON_PROMOTIONAL_SUBSCRIPTION` and no tag for now?

nicpottier: I think that's right yes, if we don't have a reply then we continue to use the old one. (it may fail but at least we tried)

rowanseymour: Updated so it's

 1. If it's a response, use message type = `RESPONSE`, no tag
 2. If there's a topic, use message type = `MESSAGE_TAG`, with tag
 3. Use message type = `NON_PROMOTIONAL_SUBSCRIPTION`, no tag

We can revisit nearer to Jan 15th

PR #267, opened 2019-11-27
Convert to use Github actions

norkans7:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  test:
+    name: Test
+    strategy:
+      matrix:
+        pg-version: ['9.6', '10', '11']
```
rowanseymour: how come we're still doing 9.6 here?

backends/rapidpro/backend_test.go
```diff
@@ -65,6 +65,13 @@ func (ts *BackendTestSuite) SetupSuite() {
 		log.Fatalf("unable to start backend for testing: %v", err)
 	}
 
+	// read our schema sql
+	sqlSchema, err := ioutil.ReadFile("schema.sql")
```
rowanseymour: FYI the postgres action actually has an option by where you provide a directory of SQL scripts to be run when database is being initialized.. but couldn't get it to work when I was converting mailroom..

.github/workflows/ci.yml
norkans7: Can remove that, I had assumed we still want to support that for the time being

backends/rapidpro/backend_test.go
norkans7: I had the same issue as well

.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Install PostgreSQL
+      uses: harmon758/postgresql-action@v1
+      with:
+        postgresql version: ${{ matrix.pg-version }}
+        postgresql db: courier_test
+        postgresql user: courier
+        postgresql password: courier
+
+    - name: Install Linux packages
+      run: sudo apt install -y --no-install-recommends postgresql-client
```
rowanseymour: if you're not doing the sql loading in a step, then you don't need this

norkans7: right

PR #275, opened 2020-01-02
Enable quick replies on external channel

allancslima:
Enabling the quick replies of message on external channel by replacing the variable `{{quick_replies}}`. The quick replies are sent on last part of message.

Available formats:

1 - `["One", "Two", "Three"]`;
2 - `<item>One</item><item>Two</item><item>Three</item>`;
3 - `"One\,Two\,Three"` ("\\," as separator);

When channel is configured with:

- `POST/PUT` and content type `JSON`, use format 1;
- `POST/PUT` and content type `XML`, use format 2;
- else, use format 3, that's `GET` requests or content type `URL encoded`.

I'll also open a PR on the RP backend, specifying the quick replies format on example according to the send method and the content type.


nicpottier:
What's the use case here? I would rather add a specific channel here instead of jamming this into an external channel.

matmsa27:
Hi @nicpottier , all channels that use the external api, that's not supported today. For example, the use case where we want to use quick replies together with text of a sms-based external API, or if we're connecting to an IP channel that support this kind of feature. It's one more flexibility that we give for this channel that can be used for many different purposes.

We have a microservice that has communication with Rapidpro by external API, and it sends the message to a widget like [Chat Widget Sample](https://github.com/livechat/chat-widget-sample).

nicpottier:
The external channel really should only be used for one off integrations where nobody would ever want to use that channel type again. Examples may include a single contractor setting up a one time integration.

Everything else should have its own channel type so that the effort can be reused by others. IE, if you are ever using the same configuration for an external channel then at that point you should be making a new channel type.

matmsa27:
I agree with you in terms of concept of External API, but I don't understand why not enabling quick replies for those cases that we need to use it for one off integrations.

nicpottier:
Because that would encourage the wrong behavior of hacked in channels via external APIs instead of creating reusable components.

What is the requirement for this particular thing being available in external channels right now? Is that channel really a one-off that will never be used again by anybody?

matmsa27:
Hi @nicpottier , so, we have a customer that deserves to use the Live Chat on your official homepage. For this particular case, we need to configure the external API channel and send the quick-replies to show on Live Chat. This is our specific case, it's the only customer will be using this way, therefore it wouldn't be worth it to create a new channel for a specific case.

nicpottier:
That sounds like a useful thing for more than that one customer, is that backend available somewhere so we could all standardize on it?

handlers/external/external.go
```diff
@@ -300,12 +300,24 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+
 		// if we are smart, first try to convert to GSM7 chars
 		if encoding == encodingSmart {
 			replaced := gsm7.ReplaceSubstitutions(part)
 			if gsm7.IsValid(replaced) {
 				form["text"] = replaced
 			}
+
+			if len(msg.QuickReplies()) > 0 {
+				replaced := gsm7.ReplaceSubstitutions(form["quick_replies"])
```
nicpottier: Not sure this makes any sense to include. Quick replies really aren't going to be used somewhere that only support GSM7 encoding.

handlers/external/external.go
```diff
@@ -344,13 +356,41 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+}
+
 func replaceVariables(text string, variables map[string]string, contentType string) string {
 	for k, v := range variables {
 		// encode according to our content type
 		switch contentType {
 		case contentJSON:
-			marshalled, _ := json.Marshal(v)
-			v = string(marshalled)
+			if !(len(v) > 1 && v[0] == '[' && v[len(v) - 1] == ']') {
```
nicpottier: This is a bit hacky and almost certainly going to blow up at some point. Better would be to pass in whether the thing being passed in needs to be escaped or not and skip over this in the case where it isn't. (IE: escaping needs to take place in `buildQuickReplies` then that value is passed to this with escaping off)

handlers/external/external.go
```diff
@@ -344,13 +356,41 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+		separator = "</item><item>"
+
+	default:
+		response = ""
+		format = "%s"
+		separator = "\\,"
+	}
+
+	if len(quickReplies) > 0 {
+		response = fmt.Sprintf(format, strings.Join(quickReplies, separator))
```
nicpottier: I don't think this works in all cases. IE, if a quick reply has `/>` in it then that wrecks the format, same goes for one that contains a `"` in a JSON reply.

The quick reply values need to be escaped appropriately according to the content type.

handlers/external/external.go
```diff
@@ -344,13 +356,41 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+
+	case (sendMethod == http.MethodPost || sendMethod == http.MethodPut) && contentType == contentXML:
+		response = ""
+		format = "<item>%s</item>"
+		separator = "</item><item>"
+
+	default:
+		response = ""
+		format = "%s"
+		separator = "\\,"
```
rowanseymour: is there a more natural separator than `\,`? just thinking that `x\,y` is a valid quick reply anyway

nicpottier: Good point, maybe the appropriate way to do this is just have repeating values, ie:
  `text=foo&quick_reply=one&quick_reply=two`

rowanseymour: yeah was thinking that tho it means `{{ quick_replies }}` contains the key/values so you can't say so `qr=one&qr=two` but maybe that's ok

allancslima: I submitted some changes, but I kept the format with `\,`, what do you decide?

allancslima: And how would it be replaced in the send URL/body? Because the default is key-value pairs.

rowanseymour: The tricky part is that we don't restrict any characters in quick replies.. so no character(s) is strictly safe as a separator. So I think I'd lean toward it evaluating like...

`http://endpoint.com?{{quick_replies}}` 

=>

`http://endpoint.com?quick_reply=one&quick_reply=two`

Since that's the most correct way to send multiple values in a URL

allancslima: Great! And for the tests, by default is it allowed to compare multiple values ​​this way?

handlers/external/external.go
```diff
@@ -344,7 +356,34 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 	return status, nil
 }
 
-func replaceVariables(text string, variables map[string]string, contentType string) string {
+func buildQuickRepliesResponse(quickReplies []string, sendMethod string, contentType string) string {
+	if quickReplies == nil {
+		quickReplies = []string{}
+	}
+	switch {
+	case (sendMethod == http.MethodPost || sendMethod == http.MethodPut) && contentType == contentJSON:
```
nicpottier: this isn't really a switch, can you change this to if/else statements?

nicpottier: this isn't really a switch, can you change this to if/else statements?

handlers/external/external.go
rowanseymour: not sure I understand... tests? 

@nicpottier how you feel about this?

allancslima: Yes, at `external_test.go`.

handlers/external/external.go
```diff
@@ -344,7 +356,40 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+		for _, reply := range quickReplies {
+			buf := &bytes.Buffer{}
+			err := xml.EscapeText(buf, []byte(reply))
+
+			if err == nil {
+				reply = buf.String()
+				response.WriteString(fmt.Sprintf("<item>%s</item>", reply))
+			}
+		}
+		return response.String()
```
nicpottier: Can't you just use `xml.Marshal` here with an appropriately named struct?

allancslima: Do you mean create a struct like the following?

```
type foo struct {
  items []string `xml:item`
}
```

So, the `xml.Marshal` could return something like:

```
"<foo><item>a</item><item>b</item><item>c</item></foo>"
```

Still needed to remove the `foo` tag, alright?

nicpottier: I think I'd be alright even with just concatenating the single `<item>foo</item>` elements together from individual `xml.Marshal`s

matmsa27:
Hey @nicpottier any news?

nicpottier:
Thanks!

PR #277, opened 2020-01-11
Hotfix FCM quick replies

johncordeiro:
This is a hotfix in the FCM quick replies that first is sending quick replies outside the `data` node in the JSON which is wrong according to the firebase [docs](https://firebase.google.com/docs/cloud-messaging/concept-options#data_messages), second to send more a simpler structure with a string array instead of an object and third to only send quick replies in the last part of the msg.


johncordeiro:
any comment on this @nicpottier ?

nicpottier:
thanks!

PR #284, opened 2020-03-13
Strip cookie from incoming requests

nicpottier:
Little cleanup here and there too.


courier.toml
```diff
@@ -16,7 +16,7 @@ backend = "rapidpro"
 port = 8080
 
 # Our database connection string, right now only Postgres is supported
-db = "postgres://courier@localhost/courier?sslmode=disable"
+db = "postgres://temba:temba@localhost/temba?sslmode=disable"
```
nicpottier: These match default mailroom / dev settings now.

server.go
```diff
@@ -286,6 +286,9 @@ func (s *server) channelHandleWrapper(handler ChannelHandler, handlerFunc Channe
 
 		// read the bytes from our body so we can create a channel log for this request
 		response := &bytes.Buffer{}
+
+		// Trim out cookie header, should never be part of authentication and can leak auth to channel logs
+		r.Header.Del("Cookie")
```
rowanseymour: any argument for redacting this instead of removing it?

nicpottier: I thought about it but that was 3 lines of code instead of one so figured why bother. ;)

PR #286, opened 2020-03-17
Enhancement/wpp contact check

koallann:
Sending this PR to the demand that emerged in issue [#1169](https://github.com/rapidpro/rapidpro/issues/1169), as proposed by [rudigiesler](https://github.com/rudigiesler):

> 1. Try do a POST /v1/messages.
> 2. If it fails with the specific contact error, then do a contact check by POST /v1/contacts
> 3. If the contact check is valid, do the POST /v1/messages again.

### DIFF
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -22,6 +22,9 @@ import (
 const (
 	configNamespace = "fb_namespace"
 )
+var (
+	retryParam = ""
+)
 
 func init() {
 	courier.RegisterHandler(newHandler())
@@ -547,9 +550,10 @@ func sendWhatsAppMsg(msg courier.Msg, sendPath *url.URL, token string, payload i
 	rr, err := utils.MakeHTTPRequest(req)
 	log := courier.NewChannelLogFromRR("Message Sent", msg.Channel(), msg.ID(), rr).WithError("Message Send Error", err)
 	errPayload := &mtErrorPayload{}
+	err = json.Unmarshal(rr.Body, errPayload)
 
 	// handle send msg errors
-	if err = json.Unmarshal(rr.Body, errPayload); err == nil && len(errPayload.Errors) > 0 {
+	if err == nil && len(errPayload.Errors) > 0 {
 		if !hasWhatsAppContactError(*errPayload) {
 			err := errors.Errorf("received error from send endpoint: %s", errPayload.Errors[0].Title)
 			return "", []*courier.ChannelLog{log}, err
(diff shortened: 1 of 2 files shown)


handlers/whatsapp/whatsapp.go
```diff
@@ -523,38 +533,119 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+	req, _ := http.NewRequest(http.MethodPost, sendPath.String(), bytes.NewReader(jsonBody))
+	req.Header = buildWhatsAppRequestHeader(token)
 	rr, err := utils.MakeHTTPRequest(req)
 	log := courier.NewChannelLogFromRR("Message Sent", msg.Channel(), msg.ID(), rr).WithError("Message Send Error", err)
-	if err != nil {
-		return "", log, err
+	errPayload := &mtErrorPayload{}
+
+	// handle send msg errors
+	if err = json.Unmarshal(rr.Body, errPayload); err == nil && len(errPayload.Errors) > 0 {
```
nicpottier: Please unpack this. For the super simple cases this construct is ok but for anything more than the simple != `nil` case it should be exploded.

handlers/whatsapp/whatsapp_test.go
```diff
@@ -380,6 +380,60 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
+				Method: "POST",
+				Path:   "/v1/messages",
+				Body:   `{"to":"250788123123","type":"text","text":{"body":"try again"}}`,
+			}: MockedResponse{
+				Status: 201,
+				Body:   `{"messages": [{"id": "157b5e14568e8"}]}`,
+			},
+		},
+		SendPrep: setSendURL,
+	},*/
```
nicpottier: Why commented out?

koallann: Oh sorry! I've added a review comment that you can't see, I think...

Because I have the same key twice the last key will overlap the first, so I cannot reproduce the messaging retry after contact check. Can I resolve this by putting a query param in the messaging request? Like `?retry=1`.

handlers/whatsapp/whatsapp.go
koallann: Hmm... doesn't the `len(errPayload.Errors) > 0` statement ensure that the errors will be handled safely?

nicpottier: All I'm saying here is unpack it.. ie:

```golang
err = json.Unmarshal(rr.Body, errPayload);
if err == nil && len(errPayload.Errors) > 0 {
   ...
```

handlers/whatsapp/whatsapp_test.go
nicpottier: Ya, that could work, you could even make that a variable that is set in the unit test only.(ie a private var `retryParam` that is by default `""` but you set in the unit test so that you can differentiate these.

koallann: Great!

PR #287, opened 2020-03-24
Retry WhatsApp messaging after contact verification with returned WhatsApp ID

koallann:
PR https://github.com/nyaruka/courier/pull/286 makes contact verification by `POST /v1/contacts`, validating the phone number before a new retry to `POST /v1/messages`, when the first try fails with error `"unknown contact"`. Response body format for contact verification:
```json
{
  "contacts": [{"input": "16315551000", "status": "valid", "wa_id": "16315551000"}]
}
```
This PR adjusts the retry request to use the `"wa_id"` returned from contact verification instead of the URN, because we may have inconsistencies in the format of phone numbers in some country when they are registered via API or manually.

In our case, in Brazil, a 9th digit has been incorporated into phone numbers in recent years, but not in WhatsApp, there are regions with 8-digit phone numbers while they are registered in the 9-digit pattern on RP, for example, "+55 98877-6655" is the current phone number pattern, while in WhatsApp it is "+55 8877-6655", causing problems with the new attempt to send message. Contact verification with "+55 98877-6655" or "+55 98877-6655" would return the same `"wa_id"`, which is "5588776655".

So, this PR makes messaging more reliable in these cases of phone number inconsistencies with WhatsApp.

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -306,9 +306,11 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus) 
 	timeout, cancel := context.WithTimeout(ctx, backendTimeout)
 	defer cancel()
 
-	err := b.handleContactURNUpdate(ctx, status)
-	if err != nil {
-		return errors.Wrap(err, "error updating contact URN")
+	if status.HasUpdatedURN() {
+		err := b.updateContactURN(ctx, status)
+		if err != nil {
+			return errors.Wrap(err, "error updating contact URN")
+		}
 	}
 	// if we have an ID, we can have our batch commit for us
 	if status.ID() != courier.NilMsgID {
@@ -341,18 +343,9 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus) 
 	return nil
 }
 
-// handleContactURNUpdate handles updating contact URN according to the old URN and the new URN from status
-func (b *backend) handleContactURNUpdate(ctx context.Context, status courier.MsgStatus) error {
-	oldURN, newURN := status.UpdatedURN()
-
-	// both URNs must have the same scheme
-	if oldURN.Scheme() != newURN.Scheme() {
-		return errors.New("cannot update contact URN to a different URN scheme")
-	}
-	// ignore update if we have the same URN path
-	if oldURN.Path() == newURN.Path() {
-		return nil
-	}
+// updateContactURN updates contact URN according to the old/new URNs from status
+func (b *backend) updateContactURN(ctx context.Context, status courier.MsgStatus) error {
+	old, new := status.UpdatedURN()
 
 	// retrieve channel
 	channel, err := b.GetChannel(ctx, courier.AnyChannelType, status.ChannelUUID())
@@ -364,18 +357,18 @@ func (b *backend) handleContactURNUpdate(ctx context.Context, status courier.Msg
 	if err != nil {
 		return err
 	}
-	// retrieve the oldURN URN
-	oldContactURN, err := selectContactURN(tx, dbChannel.OrgID(), oldURN)
+	// retrieve the old URN
+	oldContactURN, err := selectContactURN(tx, dbChannel.OrgID(), old)
 	if err != nil {
 		return errors.Wrap(err, "error retrieving old contact URN")
 	}
-	// retrieve the newURN URN
-	newContactURN, err := selectContactURN(tx, dbChannel.OrgID(), newURN)
+	// retrieve the new URN
+	newContactURN, err := selectContactURN(tx, dbChannel.OrgID(), new)
 	if err != nil {
-		// only update the oldURN URN path if the newURN URN doesn't exist
+		// only update the old URN path if the new URN doesn't exist
 		if err == sql.ErrNoRows {
-			oldContactURN.Path = newURN.Path()
-			oldContactURN.Identity = string(newURN.Identity())
+			oldContactURN.Path = new.Path()
+			oldContactURN.Identity = string(new.Identity())
 
 			err = fullyUpdateContactURN(tx, oldContactURN)
 			if err != nil {
@@ -387,11 +380,11 @@ func (b *backend) handleContactURNUpdate(ctx context.Context, status courier.Msg
 		return errors.Wrap(err, "error retrieving new contact URN")
 	}
 
-	// only update the newURN URN if it doesn't have an associated contact
+	// only update the new URN if it doesn't have an associated contact
 	if newContactURN.ContactID == NilContactID {
 		newContactURN.ContactID = oldContactURN.ContactID
 	}
-	// remove contact association from oldURN URN
+	// remove contact association from old URN
 	oldContactURN.ContactID = NilContactID
 
 	// update URNs
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -308,7 +308,7 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus) 
 
 	err := b.handleContactURNUpdate(ctx, status)
 	if err != nil {
-		return errors.Wrap(err, "error on update contact URN")
+		return errors.Wrap(err, "error updating contact URN")
 	}
 	// if we have an ID, we can have our batch commit for us
 	if status.ID() != courier.NilMsgID {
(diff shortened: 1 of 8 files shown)


nicpottier:
If we aren't updating the URN won't this cause every single message sent to that URN to go through the process of attempted send, lookup, then resend?

koallann:
Yes, the idea is to keep the format of the registered phone number, to facilitate a data export, for example. In our case, phone calls and SMS only works with 9 digits and I also believe that WhatsApp can still incorporate the 9th digit into all `"wa_id"`.

nicpottier:
But doesn't that mean that every send to those numbers is then three calls?

nicpottier:
Also for incoming messages, won't they get mapped to a new contact if the URN isn't updated?

koallann:
Good point! I'll update this soon.

handlers/whatsapp/whatsapp.go
```diff
@@ -573,6 +573,52 @@ func sendWhatsAppMsg(msg courier.Msg, sendPath *url.URL, token string, payload i
...
+		wppID, err := jsonparser.GetString(rrCheck.Body, "contacts", "[0]", "wa_id")
+
+		if err == nil {
+			// retrieve contact
+			urn, _ := urns.NewWhatsAppURN(msg.URN().Path())
+			contact, _ := h.Backend().GetContact(ctx, msg.Channel(), urn, "", "")
+			// update URN
+			h.Backend().RemoveURNfromContact(ctx, msg.Channel(), contact, urn)
+			urn, _ = urns.NewWhatsAppURN(wppID)
+			h.Backend().AddURNtoContact(ctx, msg.Channel(), contact, urn)
```
koallann: There is no an update function?

koallann:
Shouldn't this URN update be in contact creation model? Makes sense?

nicpottier:
What do you mean by contact creation mode?

I think in this case we would want to add a method to `MsgStatus`, say `SetUpdatedURN(urn urns.URN)` and `UpdatedURN() urns.URN`. Channels that find they have updated the URN during sending (as in this case) would call the `SetUpdatedURN` method and it is then up to the backend to update that.

RapidPro backend would not do a batch commit in this case, but instead first update the URN path to the new path, then commit the msg status. (come to think of it it could probably do the URN update first, then if eligible add the status to be batch committed)

nicpottier:
One thing to make sure to account for is if the "updated" URN already exists.. 

1. If that URN already exists but with no associated contact, then we associate our old contact to it and disassociate our old URN
2. If the URN already exists and has an active different contact assigned to it, then that's trickier, we can't merge the message history there. So maybe we just deactivate the old URN? (replies will go to the new contact which is confusing but what would occur anyways, garbage in, garbage out)
3. Simplest case is if the URN doesn't exist, then we can just update the path on the current URN

koallann:
Thanks for the explanations @nicpottier! Validate this latest implementation, please, tests are pending for now...

test.go
```diff
@@ -575,6 +577,11 @@ func (m *mockMsgStatus) ChannelUUID() ChannelUUID { return m.channel.UUID() }
 func (m *mockMsgStatus) ID() MsgID                { return m.id }
 func (m *mockMsgStatus) EventID() int64           { return int64(m.id) }
 
+func (m *mockMsgStatus) OldURN() urns.URN       { return m.oldURN }
+func (m *mockMsgStatus) SetOldURN(urn urns.URN) { m.oldURN = urn }
+func (m *mockMsgStatus) NewURN() urns.URN       { return m.newURN }
+func (m *mockMsgStatus) SetNewURN(urn urns.URN) { m.newURN = urn }
```
nicpottier: Since old and new really go together it probably makes sense for these to be more joined.

IE:
  `SetUpdatedURN(old urns.URN, new urns.URN) error`
  `UpdatedURN() (urns.URN, urns.URN)`
  `HasUpdatedURN() bool`

backends/rapidpro/backend.go
```diff
@@ -305,6 +306,10 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus)
 	timeout, cancel := context.WithTimeout(ctx, backendTimeout)
 	defer cancel()
 
+	err := b.handleContactURNUpdate(ctx, status)
```
nicpottier: `if status.HasUpdatedURN() {
    ...
}`

backends/rapidpro/backend.go
```diff
@@ -336,6 +341,71 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus)
 	return nil
 }
 
+// handleContactURNUpdate handles updating contact URN according to the old URN and the new URN from status
+func (b *backend) handleContactURNUpdate(ctx context.Context, status courier.MsgStatus) error {
+	// both URNs must have the same scheme
```
nicpottier: `old, new := status.UpdatedURN()`

backends/rapidpro/backend.go
```diff
@@ -336,6 +341,71 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus)
...
+	if status.OldURN().Scheme() != status.NewURN().Scheme() {
+		return errors.New("cannot update contact URN to a different URN scheme")
+	}
+	// ignore update if we have the same URN path
+	if status.OldURN().Path() == status.NewURN().Path() {
+		return nil
+	}
```
nicpottier: These errors / checks can be in `SetUpdatedURN()` that way the error happens closer to where the mistake is made.

backends/rapidpro/backend.go
```diff
@@ -336,6 +341,71 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus)
...
+	dbChannel := channel.(*DBChannel)
+	tx, err := b.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	// retrieve the old URN
+	oldContactURN, err := selectContactURN(tx, dbChannel.OrgID(), status.OldURN())
+	if err != nil {
+		return err
```
nicpottier: Wrap the errors in this function to aid in tracing errors.

backends/rapidpro/urn.go
```diff
@@ -248,6 +259,8 @@ UPDATE
 SET 
 	channel_id = :channel_id, 
 	contact_id = :contact_id, 
+	identity = :identity, 
+	path = :path,
```
nicpottier: This change means we need to make sure all places that call `updateContactURN` are passing fully populated URN objects. Looking around it looks like `setDefaultURN` calls `contactURNsForContact` which doesn't select `path`. So need to update that.

backends/rapidpro/backend.go
nicpottier: Here gate with `hasUpdatedURN`

backends/rapidpro/backend.go
nicpottier: These checks should be in `SetUpdatedURN` so that any error happens closer to the place where the mistake was made.

backends/rapidpro/backend.go
```diff
@@ -336,6 +341,71 @@ func (b *backend) WriteMsgStatus(ctx context.Context, status courier.MsgStatus)
...
+func (b *backend) handleContactURNUpdate(ctx context.Context, status courier.MsgStatus) error {
+	// both URNs must have the same scheme
+	if status.OldURN().Scheme() != status.NewURN().Scheme() {
+		return errors.New("cannot update contact URN to a different URN scheme")
+	}
+	// ignore update if we have the same URN path
+	if status.OldURN().Path() == status.NewURN().Path() {
+		return nil
+	}
+
```
nicpottier: `oldURN, newURN := status.UpdatedURN()`

backends/rapidpro/backend.go
nicpottier: Should wrap all these errors so we have a bit more context if things go sideways.

backends/rapidpro/backend.go
koallann: But the logic will be duplicated to `DBMsgStatus` and `mockMsgStatus`.

nicpottier: `mockMsgStatus` doesn't need these checks at all, that's just for testing. As I said, much better to throw errors closest to where they occur, this would all be delayed.

koallann:
Hey @nicpottier any more updates here?

handlers/whatsapp/whatsapp.go
```diff
@@ -611,12 +623,53 @@ func sendWhatsAppMsg(msg courier.Msg, sendPath *url.URL, token string, payload i
...
+			return "", "", []*courier.ChannelLog{log, checkLog}, err
+		}
+		// update contact URN and msg destiny with returned wpp id
+		wppID, err := jsonparser.GetString(rrCheck.Body, "contacts", "[0]", "wa_id")
+
+		if err == nil {
+			var updatedPayload interface{}
+
+			// handle msg type casting
+			if text, ok := payload.(mtTextPayload); ok {
```
nicpottier, line 640: I think this is probably best done with a type switch: 
  https://tour.golang.org/methods/16

backends/rapidpro/urn.go
nicpottier: Was this addressed?

koallann:
Yep, we've been running this since for 2 weeks.

backends/rapidpro/urn.go
koallann: I created a separated SQL ([R268](https://github.com/nyaruka/courier/pull/287/files#diff-01ee8537f8a93225f573ebcc5fd7ea4eR268)) for this update that's be used by a separeted function ([R299](https://github.com/nyaruka/courier/pull/287/files#diff-01ee8537f8a93225f573ebcc5fd7ea4eR299)).

nicpottier:
Thanks guys!

PR #288, opened 2020-04-07
Fix sending WA template messages

norkans7:


norkans7:
All previous WA client version should now be at least v2.25.4 whose support also ends next month
v2.25.3 should not be working as of 3 days ago

https://developers.facebook.com/docs/whatsapp/changelog

nicpottier:
Praekelt has a "look alike" client that they use which I believe still uses the old format and is an active customer on TextIt. @smn does your endpoint support the new message template formats?

handlers/whatsapp/whatsapp.go
```diff
@@ -503,24 +504,41 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 
-			payload := &templatePayload{
-				To:   msg.URN().Path(),
-				Type: "template",
-			}
-			payload.Template.Namespace = namespace
-			payload.Template.Name = templating.Template.Name
-			payload.Template.Language.Policy = "deterministic"
-			payload.Template.Language.Code = templating.Language
+			hsmSupport := msg.Channel().StringConfigForKey(configHSMSupport, "")
```
nicpottier, line 507: You can use `BoolConfigForKey` and use normal JSON `bool` here.

PR #290, opened 2020-04-20
Default to channel config max length

norkans7:


handlers/arabiacell/arabiacell.go
```diff
@@ -77,7 +77,8 @@ func (h *handler) SendMsg(_ context.Context, msg courier.Msg) (courier.MsgStatus
 	}
 
 	status := h.Backend().NewMsgStatusForID(msg.Channel(), msg.ID(), courier.MsgErrored)
-	for _, part := range handlers.SplitMsg(handlers.GetTextAndAttachments(msg), maxMsgLength) {
+	maxLength := msg.Channel().IntConfigForKey(courier.ConfigMaxLength, maxMsgLength)
+	for _, part := range handlers.SplitMsg(handlers.GetTextAndAttachments(msg), maxLength) {
```
nicpottier, line 81: Doesn't `msg` have a reference already to the channel? If so, can't `GetTextAndAttachments` just do this internally maybe? Or are there parts where that isn't appropriate?

norkans7: the length is passed to `handlers.SplitMsg` 

https://github.com/nyaruka/courier/blob/f35f288cde69114d27bf665299356f2e2d86b90f/handlers/utils.go#L92

nicpottier: Would it maybe make sense to pass in the channel there then? Would reduce the times this is repeated everywhere.

norkans7: That will be complicated for https://github.com/nyaruka/courier/pull/290/files#diff-2bf19872ed21c9bb1a9b493c01920ac9R355 since these we have the max for a message and the max for a description

PR #295, opened 2020-05-04
Allow to use config URL if set

norkans7:


handlers/telesom/telesom.go
```diff
@@ -69,17 +69,22 @@ func (h *handler) receiveMessage(ctx context.Context, channel courier.Channel, w
...
+		return nil, fmt.Errorf("no password set for TS channel")
 	}
 
 	privateKey := msg.Channel().StringConfigForKey(courier.ConfigSecret, "")
 	if privateKey == "" {
-		return nil, fmt.Errorf("no private key set for Telesom channel")
+		return nil, fmt.Errorf("no private key set for TS channel")
+	}
+
+	tsSendURL := msg.Channel().StringConfigForKey(courier.ConfigSendURL, "")
```
nicpottier, line 85: can just be: 
  `msg.Channel().StringConfigForKey(courier.ConfigSendURL, sendURL)` no?

norkans7: I was thinking that would miss if "" is set in the config but we can make sure the URL is required on the claim form

nicpottier:
Oh, well I guess depends on whether we think from now on that always has to be set or not.

Issue #296, opened 2020-05-05
Sending limit on the LINE channel


koallann:
The LINE channel handler sends messages to their contacts though "push" action, there are two sending types:

`Reply` - response to an event from a user, group, or room, requires a `replyToken`, token of the last user message.
`Push` - message to a user, group, or room at any time.

https://developers.line.biz/en/reference/messaging-api/#send-reply-message
https://developers.line.biz/en/reference/messaging-api/#send-push-message

`Reply` messages are free, `Push` has a limit of up to 500 messages per month:
![image](https://user-images.githubusercontent.com/24883556/81063749-ccd22800-8eae-11ea-8c99-03f54e3ddf48.png)
(https://www.linebiz.com/jp-en/other/)

We would like to implement this sending treatment to decrease extensively the financial cost of those who use this channel, but before modeling a solution, we want to know if it's allowed to make this type of change in the RapidPro platform, that's, to involve the Django app adding new db models to attend this handler specific restrictions.

nicpottier:
This seems sane enough, what do you expect needs to change? We already have information on outgoing messages as to whether they are replies, if it is just a matter of changing the endpoint that is being called and some extra parameters then it may just be a matter of updating the Line handler on the Courier side.

koallann:
I noticed the `Msg.ResponseToExternalID()` and it looks enough, that's, just set the `replyToken` as `ExternalID` for incoming messages.

Example webhook msg object:

```json
{
  "destination": "xxxxxxxxxx",
  "events": [
    {
      "replyToken": "0f3779fba3b349968c5d07db31eab56f",
      "type": "message",
      "mode": "active",
      "timestamp": 1462629479859,
      "source": {
        "type": "user",
        "userId": "U4af4980629..."
      },
      "message": {
        "id": "325708",
        "type": "text",
        "text": "Hello, world"
      }
    }
  ]
}
```
So, on `SendMsg`, just check that `msg.ResponseToExternalID` is not empty and send message to `/v2/bot/message/reply`.

Example request object:

```json
{
  "replyToken":"nHuyWiB7yP5Zw52FIkcQobQuGDXCTA",
  "messages":[
    {
      "type":"text",
      "text":"Hello, user"
    },
    {
      "type":"text",
      "text":"May I help you?"
    }
  ]
}
```

Sounds right?

nicpottier:
Yep, that looks light it would work!

koallann:
Great! Many thanks @nicpottier

PR #299, opened 2020-05-11
Enhancement/LINE sending

koallann:
https://github.com/nyaruka/courier/issues/296

### DIFF
--- a/handlers/line/line.go
+++ b/handlers/line/line.go
@@ -239,21 +239,32 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 		batchCount++
 
 		if batchCount == maxMsgSend || (i == len(jsonMsgs)-1) {
-			if req, err := buildSendMsgRequest(authToken, msg.URN().Path(), msg.ResponseToExternalID(), batch); err == nil {
-				rr, err := utils.MakeHTTPRequest(req)
-				log := courier.NewChannelLogFromRR("Message Sent", msg.Channel(), msg.ID(), rr).WithError("Message Send Error", err)
-				status.AddLog(log)
+			req, err := buildSendMsgRequest(authToken, msg.URN().Path(), msg.ResponseToExternalID(), batch)
+			if err != nil {
+				return status, err
+			}
+			rr, err := utils.MakeHTTPRequest(req)
+			log := courier.NewChannelLogFromRR("Message Sent", msg.Channel(), msg.ID(), rr).WithError("Message Send Error", err)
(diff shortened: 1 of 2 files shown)


handlers/line/line.go
```diff
@@ -236,12 +239,17 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 		if batchCount == maxMsgSend || (i == len(jsonMsgs)-1) {
-			if req, err := buildSendMsgRequest(authToken, msg.URN().Path(), batch); err == nil {
+			if req, err := buildSendMsgRequest(authToken, msg.URN().Path(), msg.ResponseToExternalID(), batch); err == nil {
 				rr, err := utils.MakeHTTPRequest(req)
 				log := courier.NewChannelLogFromRR("Message Sent", msg.Channel(), msg.ID(), rr).WithError("Message Send Error", err)
 				status.AddLog(log)
 
 				if err != nil {
+					msg, err := jsonparser.GetString(rr.Body, "message")
+					// don't retry messages for invalid reply token
```
koallann, line 249: Because the `replyToken` must be of the last user message.

koallann:
We can put these changes in production before a merge.

handlers/line/line.go
nicpottier: Are you saying that if a user sends two messages "one" and "two".. if we use the reply token from "one" in a reply it will fail? Seems like that's a real possibility.

Should we back down to a push in this case?

koallann: Exactly. I guess that way the users will have a greater sense of security regarding the functioning of the LINE channel, they will be aware that they will be consuming their plan limits (up to 500 for free plan) in triggering messages.

nicpottier: So do you agree we should be switching to a push there or not?

koallann: Flows for social networks are very much based on user's response, right? So if an "old" message fails and a retry consumes some limit, I think we should not switch to `push '.

nicpottier: But it might still be interactive. If a contact is in a flow and responds to a question with two answer right after each other "yes", "of course", then what you are saying is the next flow question/response would fail completely because it would have a "response_to" id for the "yes" message. That seems wrong to me. Even if used sparingly it feels more right to try for a reply and if that doesn't work then use a push.

koallann: Hmm, well noted!!

koallann: @nicpottier FYI the last requested changes were made.

PR #303, opened 2020-05-20
Add FBA Channel type support

norkans7:


backends/rapidpro/channel.go
```diff
@@ -139,21 +139,143 @@ const localTTL = 60 * time.Second
...
+       ch.config as config,
+       org.config as org_config,
+       org.is_anon as org_is_anon
+FROM
+       channels_channel ch
+       JOIN orgs_org org on ch.org_id = org.id
+WHERE
+       ch.address = $1 AND
+       ch.is_active = true AND
+       ch.org_id IS NOT NULL`
```
nicpottier, line 195: Don't think we need org_id is not null anymore..

backends/rapidpro/channel.go
```diff
@@ -139,21 +139,143 @@ const localTTL = 60 * time.Second
...
+
+// ChannelForUUID attempts to look up the channel with the passed in UUID, returning it
```
nicpottier, line 197: update comment

backends/rapidpro/channel.go
```diff
@@ -139,21 +139,143 @@ const localTTL = 60 * time.Second
...
+func loadChannelByAdrressFromDB(ctx context.Context, db *sqlx.DB, channelType courier.ChannelType, address courier.ChannelAddress) (*DBChannel, error) {
```
nicpottier, line 198: typo

handlers/facebookapp/facebookapp.go
```diff
@@ -0,0 +1,691 @@
...
+
+		// this was wired successfully
+		status.SetStatus(courier.MsgWired)
+	}
+
+	return status, nil
+}
+
+// ReceiveVerify handles Facebook's webhook verification callback
+func (h *handler) DescribeURN(ctx context.Context, channel courier.Channel, urn urns.URN) (map[string]string, error) {
```
nicpottier, line 616: Comment wrong

handlers/forms.go
```diff
@@ -96,3 +93,12 @@ func DecodeAndValidateXML(envelope interface{}, r *http.Request) error {
 
 	return nil
 }
+
+// ReadBody of a HTTP request up to limit bytes and make sure the Body is not consumed
+func ReadBody(r *http.Request, limit int64) ([]byte, error) {
+	body, err := ioutil.ReadAll(io.LimitReader(r.Body, limit))
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	defer r.Body.Close()
```
nicpottier, line 101: This isn't doing anything. (NoopCloser does nothing on Close)

server.go
```diff
@@ -324,7 +318,7 @@ func (s *server) channelHandleWrapper(handler ChannelHandler, handlerFunc Channe
 		}
 
 		// if no events were created we still want to log this to the channel, do so
-		if len(events) == 0 {
+		if len(events) == 0 && channel != nil {
```
nicpottier, line 321: in what scenario do we get here without a channel but with events?

norkans7: we use the URL `/c/fba/receive` for the channel type so we will figure out the channel in the handler(not here by UUID)

nicpottier: Why isn't the channel being returned by the FBA channel in `GetChannel`?

norkans7: Right that is not needed

norkans7: I remembered why we needed this we only add channel logs when we have a channel matched
https://github.com/nyaruka/courier/pull/311

PR #305, opened 2020-05-22
Fix FB location attachment to be handled at geo attachment

norkans7:


norkans7:
Fixes https://github.com/rapidpro/rapidpro/issues/1223

handlers/facebook/facebook.go
```diff
@@ -182,7 +182,12 @@ type moPayload struct {
 				Attachments []struct {
 					Type    string `json:"type"`
 					Payload *struct {
-						URL string `json:"url"`
+						URL         string `json:"url"`
+						StickerID   int64  `json:"sticker_id"`
```
norkans7, line 186: newer API has sticker ID in the attachment so I added that here well and we'll support both cases, and the older taking over if present, so we loop on attachment before checking the stickerID outside template and we'll set the later as text in case it is present

norkans7:
WhatsApp has that already https://github.com/nyaruka/courier/blob/master/handlers/whatsapp/whatsapp.go#L183

norkans7:
Probably API has changed if you noticed that not supported, I'll check on that again

nicpottier:
Oh I think I'm actually confusing with Facebook issue I opened on this! NM

PR #308, opened 2020-06-25
Add support for more template languages on WhatsApp handler

koallann:
https://github.com/rapidpro/rapidpro/issues/1206

https://github.com/nyaruka/rapidpro/pull/2836


nicpottier:
Hey, apologies, I reverted this as I don't think it is quite right from what I can tell.

In the RapidPro PR: 
  https://github.com/nyaruka/rapidpro/pull/2836/files

It appears to me that we are setting country code on all templates, which is fine. But then in this courier code aren't we then expecting to find a match based on the country/language code? IE, for the English/US case, isn't it going to look for a mapping for `eng_US` which isn't actually present?

Seems we need a two part lookup to find the proper WhatsApp language code to use:
 * First try the country/language pair lookup, `eng_US` or `spa_BR`, if found use that.
 * If not found, use only the language code `eng` or `spa` and use that
 * If not found, then it's an error

koallann:
Not necessarily because there are also mappings for these "base" languages ​​(English, Spanish and Portuguese, specifically), as here:: https://github.com/nyaruka/courier/pull/308/files#diff-3b05f66223be3994d25cf04fda0cfca0R825

koallann:
eng -> en
spa -> es
zho -> zh_CN
por -> pt_PT

https://developers.facebook.com/docs/whatsapp/message-templates/creation?locale=en_US#translations

handlers/whatsapp/whatsapp.go
```diff
@@ -776,8 +776,12 @@ func (h *handler) getTemplate(msg courier.Msg) (*MsgTemplating, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "invalid templating definition")
 	}
+	// check country
+	if templating.Country != "" {
```
nicpottier, line 780: This is set on all templating now in RapidPro isn't it? So for a US channel, it will be set to "US"

handlers/whatsapp/whatsapp.go
```diff
@@ -776,8 +776,12 @@ func (h *handler) getTemplate(msg courier.Msg) (*MsgTemplating, error) {
...
+		templating.Language = fmt.Sprintf("%s_%s", templating.Language, templating.Country)
+	}
 
-	// map our language from iso639-3 to the WA country / iso638-2 pair
+	// map our language from iso639-3_iso3166-2 to the WA country / iso638-2 pair
 	language, found := languageMap[templating.Language]
```
nicpottier, line 785: Here we will then try to look up in our language map using `eng_US` instead of just `eng` no?

handlers/whatsapp/whatsapp.go
koallann: No, the country isn't defined by channel, it's defined by the language in WhatsApp sync: https://github.com/nyaruka/rapidpro/pull/2836/files#diff-4714c50434646dabb7e2ae03726b788dR202

Just a quick review of behavior:
- On WhatsApp sync, the language is mapped from WhatsApp format to the fields `language` (iso639-3) and `country` (iso3166-2);
- On Courier, the language is mapped back to WhatsApp format;
- Examples:
```
('eng', None) -> 'eng'    -> 'en'
('eng', 'US') -> 'eng_US' -> 'en_US'
...
```
The entry `por` on language map, for example, was not removed, it was kept pointing to `pt_PT`.

handlers/whatsapp/whatsapp.go
koallann: There is a check below this code, converting the `language` to the `language_country` format, if template has a country.

koallann: Note that the `language` will be as `eng_US` if template has a country, otherwise it will be as `eng`.

koallann: And both `eng` and `eng_US` are in the `languageMap`.

handlers/whatsapp/whatsapp.go
nicpottier: Ah, understand now, forgot country was coming from the template itself not from the channel. Ok!

PR #310, opened 2020-06-29
HX channel sends MO using ISO 8859-1 encoding

norkans7:


nicpottier:
It isn't GSM7? 8859-1 is a weird one for sure.

norkans7:
No, I think ISO 8859-1 is the one they use, also the send us multipart messages as parts however I do not think there is way we can reconstruct the entire message from many requests.

So only this is what I think we can adjust to support French accents and special characters

nicpottier:
Did they say 8859-1? That is just super weird and GSM7 looks similar.

norkans7:
a contact tested the message "je suis très satisfait " and I just tried GSM7 and that does not match the text 
only 8859-1 matches correctly the message the contact sent

PR #311, opened 2020-06-29
Only log channel events when we have a channel matched

norkans7:


norkans7:
When verifying the FBA Webohook URL we do match any channel so e cannot create a channel log

PR #318, opened 2020-07-23
Fix fb postback ref ad

norkans7:


nicpottier:
Not sure I understand, has this ever worked? what changed?

norkans7:
We were not saving the Ad ID for postback referrals which seems to be used now

PR #325, opened 2020-08-26
Cache WhatsApp Medias

koallann:
Implementation of the WhatsApp medias cache proposed by @nicpottier, discussed [here](https://github.com/rapidpro/rapidpro/issues/1310).

I would just like to highlight that even if the media upload to WhatsApp fails, the message will be delivered with the attachment link instead of a media id.


handlers/whatsapp/whatsapp.go
```diff
@@ -600,7 +605,88 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+// fetchMediaID tries to fetch the id for the uploaded media, setting the result in redis.
+func (h *handler) fetchMediaID(msg courier.Msg, mimeType, mediaURL string) (string, []*courier.ChannelLog, error) {
+	var logs []*courier.ChannelLog
+	start := time.Now()
+
+	// check on cache first
+	rc := h.Backend().RedisPool().Get()
+	defer rc.Close()
+
+	cacheKey := fmt.Sprintf("whatsapp_media:%s:%s", msg.Channel().UUID().String(), mediaURL)
```
nicpottier, line 617: Instead of having one key/value at the top level for each file why don't we group these by channel. It will be more efficient from a memory perspective for Redis and easier to grok if looking at plain keys.

So basically ah HSET with `whatsapp_media:[channel_uuid]`as the set name, `mediaURL` as the key name and the value being the id. Expiring daily.

handlers/whatsapp/whatsapp.go
```diff
@@ -600,7 +605,88 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+	baseURL := msg.Channel().StringConfigForKey(courier.ConfigBaseURL, "")
+	req, err = http.NewRequest("POST", baseURL + "/v1/media", bytes.NewReader(res.Body))
+	if err != nil {
+		return "", logs, err
+	}
+	setWhatsAppAuthHeader(&req.Header, msg.Channel())
+	req.Header.Add("Content-Type", mimeType)
+	res, err = utils.MakeHTTPRequest(req)
+	if err != nil {
+		// put on failure cache
```
nicpottier, line 657: minor `put in failure cache`

handlers/whatsapp/whatsapp.go
```diff
@@ -600,7 +605,88 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+	// take uploaded media id
+	mediaID, err = jsonparser.GetString(res.Body, "media", "[0]", "id")
+	if err != nil {
+		elapsed := time.Now().Sub(start)
+		log := courier.NewChannelLogFromError("error reading the media id from response", msg.Channel(), msg.ID(), elapsed, err)
+		logs = append(logs, log)
+		return "", logs, err
+	}
+
+	// put on cache
```
nicpottier, line 678: `put in cache`

handlers/whatsapp/whatsapp.go
nicpottier: `whatsapp_media` should be a const same with failure key.

koallann: Do you mean to expire the `whatsapp_media:[channel_uuid]`? As far as I know, Redis only supports TTL at top-level keys.

nicpottier: Ya, generally the way we do this is have the date in the key as well so that those keys "roll" every day and then expire out. The main argument for that is that it is both easier to clear out caches for a channel and that it is less memory in Redis because the rather verbose portion of the key `whatsapp_media:uuid-uuid-uuid-uuid-uuid-uuid` is not repeated for every entry.

See: 
  https://instagram-engineering.com/storing-hundreds-of-millions-of-simple-key-value-pairs-in-redis-1091ae80f74c
And:
  https://redis.io/topics/memory-optimization (Use hashes when possible)

koallann: Great, thanks for the references!

koallann:
Hey @nicpottier, any more changes here?

handlers/whatsapp/whatsapp.go
```diff
@@ -600,7 +608,127 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+		return nil, err
+	}
+
+	expiration := &mediaExpiration{}
+	err = json.Unmarshal([]byte(expirationJSON), expiration)
+	if err != nil {
+		return nil, err
+	}
+
+	if expiration.ExpiresOn.Before(time.Now()) {
```
nicpottier, line 709: What's going on here?  You are managing the expiration yourself? How are things going to get expired if  their expiration is only checked when they are fetched?

koallann: Yes, I'm managing the expiration on my own because I switched the cache value type from string to hash and I can't set an expiration for a hash field. Key-value example of this implementation:

```
"whatsapp_media:a0b118cb-b010-43d8-a0b0-5483e125a812": {
  "https://link.to/image.jpg": "{\"expires_on\": \"2020-10-10T10:00:00+00:00\", \"media_id\": \"f043afd0-f0ae-4b9c-ab3d-696fb4c8cd68\"}",
  "https://link.to/video.mp4": "{\"expires_on\": \"2020-10-15T10:00:00+00:00\", \"media_id\": \"b4c8cd68-f0ae-4b9c-ab3d-696ff043afd0\"}"
}
```

Is there any way to expire a hash field automatically?

nicpottier: That's just going to cause the cache to grow unbounded, things will never clear if they are only cleared when accessed after expiration. (simplest case of a piece of media being sent once and only once being an example)

Please use the new `rcache` stuff in `gocommon` > `v1.6.0`, that'll do what we want.

koallann: Hey @nicpottier do you recommend some specific version of Redis? I'm facing the following error when I try to get a non-cached key: `NOSCRIPT No matching script. Please use EVAL.` (github.com/garyburd/redigo)

nicpottier: Hrmm, that sounds weird, why don't you check in what you have and I'll take a look?

koallann: Look:

image

I'm using Redis 5.0.7.

nicpottier: There's nothing specific to Redis versions no, Lua is available on all versions. I'm guessing there may be something subtle going on with redigo and the caching of the Lua script in that it thinks Redis has a compiled version stored but that isn't the case. Happy to take a look if you check your stuff in but that's where I would start looking.

koallann: hmm, ok, thanks!

nicpottier: Can you check in what you have so I can take a look?

koallann: Hi @nicpottier, I still haven't been able to take a look at this, I can only confirm that I will check this week, sorry for the delay.

koallann: @nicpottier as garyburd said [here](https://github.com/gomodule/redigo/issues/354#issuecomment-412750363), we have two redis packages, `github.com/gomodule/redigo` in gocommon and `github.com/garyburd/redigo` in courier.

nicpottier: You can switch the courier dependency to `gomodule` in this PR. Should address the problem.

nicpottier (approved):
👍

PR #326, opened 2020-08-27
Discord proxy

awensaunders:
This PR allows us to use the discord proxy within courier. The bulk of the code for the discord handler here is lifted from the external handler, but is substantially simplified, as we have no need for the level of flexibility the external channel type provides. There is also a simple test suite for this channel type.


awensaunders:
This PR is not ready to be merged yet, expect a PR from the Jozian fork shortly...

PR #327, opened 2020-09-03
Discord Proxy Support

awensaunders:
This is the courier side of the changes to support discord via a proxy application. 

See [RapidPro issue 1252](https://github.com/rapidpro/rapidpro/issues/1252) for more details.

This PR adds a handler and associated test which sends/receives messages in the way the proxy application expects.


handlers/discord/discord.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/nyaruka/courier/utils"
+	"github.com/nyaruka/gocommon/urns"
+	"github.com/pkg/errors"
+)
+
+const (
+	contentURLEncoded = "urlencoded"
+	contentJSON       = "json"
+
+	configMOFromField = "mo_from_field"
```
nicpottier, line 26: Is this really a configuration item? Looking at your proxy seems these are probably always going to be static no? May as well just turn these into consts then instead of making them configurable.

handlers/discord/discord.go
```diff
@@ -0,0 +1,243 @@
...
+	configMOTextField = "mo_text_field"
+	configMODateField = "mo_date_field"
+
+	configMOResponseContentType = "mo_response_content_type"
+	configMOResponse            = "mo_response"
+)
+
+var defaultFromFields = []string{"from", "sender"}
```
nicpottier, line 34: Same here, since you are controlling both sides may as well just be specific about it.

nicpottier (approved):
Happy to merge as is if you want, but would probably remove some of the configurability just for simplicity. Also not sure if some of these callbacks are actually used on your proxy (failed/delivered for example) so if not could be simpler to remove those too.

Otherwise looks good!

awensaunders:
I've also added the quickreplies in the payload that we send to the proxy. Right now discord hasn't got any notion of quick replies, but it does have reactions (which are single characters of emoji). I believe quick replies are on the roadmap for discord, so we may as well pass them over to the proxy and not have to change courier later. 

I'll make those other changes shortly, and then I think we'll be good to merge.

handlers/discord/discord.go
awensaunders: I've made these changes. You're totally right that we don't need these to be configurable as we control both sides.

awensaunders:
Unnecessary config has been deleted. I think we should be good to merge, if it's fine by you.

rowanseymour:
@awensaunders if you can fix that conflict in go.mod we can merge this

awensaunders:
I've resolved the go.mod conflict

PR #329, opened 2020-09-04
Remove incoming message prefix for Play Mobile free accounts

koallann:
The Play Mobile service places a prefix (from a list of possible prefixes) on user messages, for free accounts.

This PR is for clearing those messages when the channel has the `incoming_prefixes` config set - the type is string array and the check is case insensitive - , so user responses can be handled normally.

### DIFF
--- a/handlers/playmobile/playmobile.go
+++ b/handlers/playmobile/playmobile.go
@@ -15,9 +15,10 @@ import (
 )
 
 const (
-	configBaseURL  = "base_url"
-	configUsername = "username"
-	configPassword = "password"
+	configBaseURL          = "base_url"
+	configUsername         = "username"
+	configPassword         = "password"
+	configIncomingPrefixes = "incoming_prefixes"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


handlers/playmobile/playmobile.go
```diff
@@ -117,6 +118,21 @@ func (h *handler) receiveMessage(ctx context.Context, c courier.Channel, w http.
...
 		}
 
+		// remove message prefix according to a list of possible prefixes, useful for free accounts
+		incomingPrefixes := c.ConfigForKey(configIncomingPrefixes, []string{})
+		if prefixes, ok := incomingPrefixes.([]string); ok {
+			for _, prefix := range prefixes {
+				text := pmMsg.Content.Text
+				prefix = strings.ToLower(prefix)
+
+				if strings.ToLower(text[:len(prefix)]) == prefix {
```
nicpottier, line 128: `strings.HasPrefix` and `strings.TrimPrefix` are here for you! :)

nicpottier (changes requested):

handlers/playmobile/playmobile.go
koallann: Ya, but we need a case-insensitive check (with space or not after the prefix).

nicpottier: `strings.HasPrefix(strings.ToLower(text), prefix)`?

Note that your current implementation will blow up for cases where text is shorter than prefix which is one argument to avoid slices where you can.

koallann: Hmm, I just preferred to convert into lowercase the possible prefix instead of all the text... really, I disregarded the case that the `incoming_prefixes` config is set and the text doesn't have a prefix, I gonna fix it.

nicpottier (approved):

PR #332, opened 2020-09-22
USSD channel

kitandara:
USSD channel implementation, more at https://github.com/rapidpro/rapidpro/issues/1298

### DIFF
--- a/backends/rapidpro/msg.go
+++ b/backends/rapidpro/msg.go
@@ -504,6 +504,7 @@ type DBMsg struct {
 	SessionID_            SessionID  `json:"session_id,omitempty"`
 	SessionTimeout_       int        `json:"session_timeout,omitempty"`
 	SessionWaitStartedOn_ *time.Time `json:"session_wait_started_on,omitempty"`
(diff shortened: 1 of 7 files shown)


nicpottier:
So.. not going to go line by line on this as the approach here isn't right. You can't maintain an in memory map of sessions in Courier because:
 1. Courier runs on multiple boxes at once for load balancing and availability, so you have no guarantee the same box that received a message will get the outgoing
 2. More importantly, Courier is meant to be ephemeral, memory of the process can't be state as it can be killed / restarted at any time.

You also can't block the web thread in Courier waiting for a response (which you have no idea will come through this same process). Even if you don't deadlock yourself you are likely to run out of file descriptors under load as you keep all these sockets open.

I think the missing piece here is that if you need synchronous behavior then you need something outside RapidPro/Courier managing that connection, that isn't something Courier can do. So Courier through a channel like this could receive the message and pass it back to that service which would then pass that response back to the caller, but RapidPro as a whole depends on everything being async so this just wouldn't work. That's also likely where your abstraction to the various USSD APIs should take place.

PR #333, opened 2020-09-23
Add session status from mailroom to MT message sent to external channel API CGI parameters

kitandara:
Mailroom passes the session status to courier. We would like to pass that on in the external API URL call.

 Relates to https://github.com/rapidpro/rapidpro/issues/1298

### DIFF
--- a/backends/rapidpro/msg.go
+++ b/backends/rapidpro/msg.go
@@ -504,7 +504,7 @@ type DBMsg struct {
 	SessionID_            SessionID  `json:"session_id,omitempty"`
 	SessionTimeout_       int        `json:"session_timeout,omitempty"`
 	SessionWaitStartedOn_ *time.Time `json:"session_wait_started_on,omitempty"`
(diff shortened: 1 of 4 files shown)


backends/rapidpro/msg.go
```diff
@@ -504,6 +504,7 @@ type DBMsg struct {
 	SessionID_            SessionID  `json:"session_id,omitempty"`
 	SessionTimeout_       int        `json:"session_timeout,omitempty"`
 	SessionWaitStartedOn_ *time.Time `json:"session_wait_started_on,omitempty"`
+	SessionStatus_ string `json:"session_status,omitempty"`
```
rowanseymour, line 507: it seems whatever editor you're using doesn't run `gofmt`.. I really recommend vscode with the go extension if you haven't tried it

kitandara: Intellij. Fixed and pushed

rowanseymour (approved):

nicpottier (approved):

PR #337, opened 2020-10-26
Receive WhatsApp contacts as a comma-separated phones text

koallann:
Maybe other contact fields should be received too. 

https://developers.facebook.com/docs/whatsapp/api/webhooks/inbound#example--contacts-message-received

### DIFF
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -129,6 +129,11 @@ type eventPayload struct {
 			MimeType string `json:"mime_type" validate:"required"`
 			Sha256   string `json:"sha256"    validate:"required"`
 		} `json:"voice"`
+		Contacts []struct{
+			Phones []struct{
+				Phone string `json:"phone"`
+			} `json:"phones"`
+		} `json:"contacts"`
 	} `json:"messages"`
 	Statuses []struct {
 		ID          string `json:"id"           validate:"required"`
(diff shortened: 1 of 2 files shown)


rowanseymour:
The other way to approach this (I think I prefer) would be to turn them into attachments. So like when we get location data from a channel, it becomes an attachment like `geo:1234567,1234567` which RapidPro can then render appropriately. Maybe these become a URN-type attachment.

nicpottier:
This does feel more like an attachment. The raw WhatsApp message does contain useful metadata such as the contact's name and that more than one phone number is attached to that contact.

Perhaps the "most right" option would be to create a VCF record for the attachment: 
 https://docs.fileformat.com/email/vcf/#:~:text=VCF%20(Virtual%20Card%20Format)%20or,create%20and%20open%20these%20files.

But that starts calling into question a lot in how a flow would process that in a useful way.

Can you tell us a bit more about the use case @koallann ? IE, how exactly is this being used in a flow, what is the user looking to do once they receive this attachment etc..

koallann:
Hi @nicpottier 

The use case is in the covid-19 pandemic context with the infected users sharing contacts of people who they have recently been in close, so health agents immediately contact these shared contacts warning about the possibility of covid contracting and what precautions to take.

paulobernardoaf:
Hey @rowanseymour @nicpottier !

We really would like to see this feature on Courier. What should be the best way to approach it based on the discussion above? 
So that we can make these changes.

rowanseymour:
Putting aside for a minute how this stuff would be used in flows, I think if RapidPro receives a message with a contact data attached, then that should be saved in its entirety without throwing away data.

[VCF/vCard](https://en.wikipedia.org/wiki/VCard) does seem to be the standard for contact data - Twilio SMS apparently supports VCF as an attachment type. There are some JSON variants (https://tools.ietf.org/html/rfc7095, https://github.com/steven-elliott/vcard-json) but nothing widely adopted.

So probably we save away a VCF file like we would an image file etc.. or possibly we encode it all directly into the attachment text. Maybe in the RapidPro UI there's an easy way to turn a VCF attachment into a new contact with one click.

But as @nicpottier points out the big question is how parsing this in a flow would work. I imagine users might want to do stuff like start the attached contact in a flow saving off some of VCF data into contact fields.

If the attachment is a remote file, do we automatically fetch that? Could we have some expression functions like:

```
@(vcf_field(input.attachments[0], "N")) -> "Jim"
@(vcf_field(input.attachments[0], "TEL", 0)) -> "(111) 555-1212"
```

Maybe if we got as far as saving VCF attachments (not a huge lift at all), that might be enough for now to let more "advanced" customers start turning those into automated actions via webhook calls etc? Whilst we figure out the challenges of working with those directly in flows.

PR #344, opened 2021-01-01
switch id to bigserial

nicpottier:


backends/rapidpro/status.go
```diff
@@ -284,7 +285,7 @@ FROM
 AS 
 	s(msg_id, channel_id, status, external_id) 
 WHERE 
-	msgs_msg.id = s.msg_id::int AND
+	msgs_msg.id = s.msg_id::bigint AND
```
rowanseymour, line 288: what happens without these casts?

rowanseymour (approved):

backends/rapidpro/status.go
nicpottier: Doesn't work sadly.. going to investigate if there's another way though.

PR #349, opened 2021-03-12
Add support for Zenvia WhatsApp

norkans7:


handlers/zenviawhatsapp/zenviawhatsapp.go
```diff
@@ -0,0 +1,267 @@
...
+	urn, err := urns.NewWhatsAppURN(payload.Message.From)
+	if err != nil {
+		return nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r, err)
+	}
+
+	contactName := payload.Visitor.Name
+
+	msgs := []courier.Msg{}
+
+	for i := 0; i < len(payload.Message.Contents); i++ {
```
nicpottier, line 111: why not `range`?

nicpottier (approved):
Do we have a number we are testing this with?

norkans7:
No, we do not have a number yet a customer wanted that so he will probably help test that as beta channel

I just followed the API docs

ntwarijoshua (approved):
LGTM. IMO, I would avoid init and use provider functions as those are easier to test

norkans7:
> LGTM. IMO, I would avoid init and use provider functions as those are easier to test

I will read up more on that and see

nicpottier:
The `init` pattern here is basically registering the type as a factory. That keeps the concerns of what types are enabled as easy as what you import in your `main`, which we have found to work well. I don't think a provider is going to do anything different there.

PR #351, opened 2021-03-29
Update TS to use POST requests

norkans7:


nicpottier:
Is there a ticket associated with this or something?

norkans7:
Yes, but I am still trying to have a message successfully sent first

norkans7:
Let's keep this as a draft until that is clear,

norkans7:
Reverted to use GET requests

Issue #353, opened 2021-04-16
WhatsApp - Excel being sent as .zip


rafaelbreno:
I've opened a ticket with WhatsApp reporting a "bug" that was converting Excel files to ".zip" when sending the document via API, the thing was, looks like courier is missing a field to define the file's name it's only setting the "caption", for example:

```go
// handlers/whatsapp.go

// Line: 110k
Document *struct {
    File     string `json:"file"      validate:"required"`
    ID       string `json:"id"        validate:"required"`
    Link     string `json:"link"`
    MimeType string `json:"mime_type" validate:"required"`
    Sha256   string `json:"sha256"    validate:"required"`
    Caption  string `json:"caption"`
    Filename string `json:"filename"` // Would add this
}

type mediaObject struct {
    ID       string `json:"id,omitempty"`
    Link     string `json:"link,omitempty"`
    Caption  string `json:"caption,omitempty"`
    Filename string `json:"filename,omitempty"`
}

// And would be possible do re-utilize the Caption to the Filename
// Line 475
if attachmentCount == 0 {
    mediaPayload.Caption = msg.Text()
    mediaPayload.Filename = msg.Text()
}
```

I think this would be one of the possible solutions, because wouldn't need to modify any other service(rapidpro/mailroom), more details about the "filename" field: [Facebook Docs](https://developers.facebook.com/docs/whatsapp/api/messages/media/).

Or even create a parser for the message text, something like:
```go
// I think this is a silly idea

// msg.Text() = "caption: UPartners list\nfilename: upartner.xlsx"
func (m *mediaObject) setCaptionFilename(msg string) {
	// Would need to have both fields "caption" and "filename"
	r := regexp.MustCompile(`(caption|filename)(:)[^\n]{1,}`)
	
	// Saving every match into []string
	strs := r.FindAllString(msg, -1)
	
	// Parsing through each string
	for _, str := range strs {
		
		// For filename Prefix
		// There's a condition, if for some reason the user decides to add 2 "filename" would only consider the first one
		if strings.HasPrefix(str, "filename:") && m.Filename == "" {
			m.Filename = strings.ReplaceAll(str, "filename: ", "")
			continue
		}
		
		// The same for Caption
		if strings.HasPrefix(str, "caption: ") && m.Caption == "" {
			m.Caption = strings.ReplaceAll(str, "caption: ", "")
			continue
		}
	}
}

// [...]
// Now implementing it into mediaPayload
// Line 475
if attachmentCount == 0 {
[...]

PR #356, opened 2021-04-27
Add Filename

rafaelbreno:
As said in #353 , I've tested locally and this is the difference:

From:
WhatsApp Image 2021-04-27 at 18 41 14

To:
WhatsApp Image 2021-04-27 at 18 41 13

### DIFF
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -114,6 +114,7 @@ type eventPayload struct {
 			MimeType string `json:"mime_type" validate:"required"`
 			Sha256   string `json:"sha256"    validate:"required"`
 			Caption  string `json:"caption"`
+			Filename string `json:"filename"`
 		} `json:"document"`
 		Image *struct {
 			File     string `json:"file"      validate:"required"`
@@ -342,9 +343,10 @@ type mtTextPayload struct {
 }
 
 type mediaObject struct {
-	ID      string `json:"id,omitempty"`
-	Link    string `json:"link,omitempty"`
-	Caption string `json:"caption,omitempty"`
+	ID       string `json:"id,omitempty"`
+	Link     string `json:"link,omitempty"`
+	Caption  string `json:"caption,omitempty"`
+	Filename string `json:"filename,omitempty"`
 }
 
 type LocalizableParam struct {
@@ -472,6 +474,7 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 				}
 				if attachmentCount == 0 {
 					mediaPayload.Caption = msg.Text()
+					mediaPayload.Filename = msg.Text()
 				}
 				payload.Document = mediaPayload
 				wppID, externalID, logs, err = sendWhatsAppMsg(msg, sendPath, payload)
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -474,7 +474,14 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 				}
 				if attachmentCount == 0 {
 					mediaPayload.Caption = msg.Text()
-					mediaPayload.Filename = msg.Text()
+				}
+				mediaPayload.Filename, err = utils.URLGetFilename(mediaURL)
+
+				logrus.WithField("channel_uuid", msg.Channel().UUID().String()).WithError(err).Error("Erro while parsing the media URL")
+
+				// Break on erroh
+				if err != nil {
+					break
 				}
 				payload.Document = mediaPayload
 				wppID, externalID, logs, err = sendWhatsAppMsg(msg, sendPath, payload)
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -477,11 +477,9 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 				}
 				mediaPayload.Filename, err = utils.BasePathForURL(mediaURL)
 
-				logrus.WithField("channel_uuid", msg.Channel().UUID().String()).WithError(err).Error("Error while parsing the media URL")
-
 				// Break out on error
 				if err != nil {
-					break
+					logrus.WithField("channel_uuid", msg.Channel().UUID().String()).WithError(err).Error("Error while parsing the media URL")
 				}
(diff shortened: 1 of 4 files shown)


nicpottier:
My problem with this is that it's this crazy hack of using the message text as a filename. Do we not know the extension from the attachment name? Can we not use a generic name instead? I really don't think I can go with having the text of a message end up being a filename.

rafaelbreno:
> My problem with this is that it's this crazy hack of using the message text as a filename. Do we not know the extension from the attachment name? Can we not use a generic name instead? I really don't think I can go with having the text of a message end up being a filename.

I know for sure that we've the `mimeType`, I even thought about creating a helper function to generate the filename using a `uuid4` for the base name, and with the help of as `map[string]string` would be possible to return the file's extension for each case, but it would be a kinda of a long, but the other option would be using the "mediaURL", I'm almost sure that it is the best option. I guess it could be done in 2 ways `regex` or a simple bytes loop, like this:
```
package main

import (
	"fmt"
	"regexp"
	"time"
)

var mediaURL string = "application/pdf:https://some-url.host.service.com/media/999/zz99/9999/da514731-4bed-428c-afb9-860dd94530cc.xlsx"

func main() {
	fmt.Println(reg())
        // da514731-4bed-428c-afb9-860dd94530cc.pdf

	fmt.Println(by())
        // da514731-4bed-428c-afb9-860dd94530cc.pdf

}

func reg() string {
	return regexp.MustCompile(`[a-zA-Z0-9\-]{1,}\.[a-z]{1,4}$`).FindString(mediaURL)
}

func by() string {
	separator := byte('/')
	for i := len(mediaURL) - 1; i >= 0; i-- {
		if mediaURL[i] == separator {
			return mediaURL[i+1:]
		}
	}
	for key, val := range mediaURL {
		fmt.Println(key, string(val), val)
	}
	return ""
}
```

nicpottier:
You shouldn't need any regexes, these are standard URLs. You can use the standard libraries to get out the path, something like:
  https://play.golang.org/p/_zw3n_TXv0s

rafaelbreno:
That's pretty good @nicpottier , I didn't know about that method, I think it should be a great option to considerate, for now I'll be doing some benchmarks, between those three, if the performance is a mutual concern, this will be a great way to define, which one to implement:
```
goos: linux
goarch: amd64
BenchmarkReg-8              136411               8506 ns/op
BenchmarkBy-8            42114620                28.5 ns/op
BenchmarkParser-8        1259732                 944 ns/op
PASS
```
The regex one is waaay slower than the others, no regex. 
Using the `by()` function is faster and maybe would need to implement an error handler for it, because I think that all mediaURL will be a URL that returns a static file, so the implemented logic should works on all cases. 
Using the solution that you wrote with the package `net/url` wouldn't need to create an error handler, but it a little slower than the `by()` method.

So Nic, what do you think?

nicpottier:
This is not something you should be benchmarking, courier is sending messages across a network so none of this matters. You should always pick the simplest, most readable and most correct implementation, which would be the URL version. (there are all sorts of edge cases which your `by()` method likely isn't dealing with (unicode, escape sequences, etc..))

rafaelbreno:
@nicpottier can you check the update? The only thing isn't clear for me is the way to treat the error, did I implemented it in the way Nyaruka is used to?

utils/misc.go
```diff
@@ -98,3 +100,12 @@ func CleanString(s string) string {
 
 	return cleaned
 }
+
+// Clean static URL, and return filename + format
+func URLGetFilename(rawURL string) (string, error) {
```
nicpottier, line 105: `BasePathForURL` is probably better.. comment should start with that as well.

handlers/whatsapp/whatsapp.go
```diff
@@ -473,6 +475,14 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 				if attachmentCount == 0 {
 					mediaPayload.Caption = msg.Text()
 				}
+				mediaPayload.Filename, err = utils.URLGetFilename(mediaURL)
+
+				logrus.WithField("channel_uuid", msg.Channel().UUID().String()).WithError(err).Error("Erro while parsing the media URL")
+
+				// Break on erroh
```
nicpottier, line 482: typo

handlers/whatsapp/whatsapp.go
```diff
@@ -473,6 +475,14 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 				if attachmentCount == 0 {
 					mediaPayload.Caption = msg.Text()
 				}
+				mediaPayload.Filename, err = utils.URLGetFilename(mediaURL)
+
+				logrus.WithField("channel_uuid", msg.Channel().UUID().String()).WithError(err).Error("Erro while parsing the media URL")
```
nicpottier, line 480: typo

nicpottier (changes requested):

nicpottier:
Couple tweaks but otherwise looks ok.

handlers/whatsapp/whatsapp.go
```diff
@@ -473,6 +475,14 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 					mediaPayload.Caption = msg.Text()
 				}
+				mediaPayload.Filename, err = utils.BasePathForURL(mediaURL)
+
+				logrus.WithField("channel_uuid", msg.Channel().UUID().String()).WithError(err).Error("Error while parsing the media URL")
+
+				// Break out on error
+				if err != nil {
+					break
+				}
```
norkans7, line 485: I think here you can create a channel log instead for the error and let the error to break on https://github.com/nyaruka/courier/pull/356/files#diff-557153ae44649c16db29083d386ef0de7914adfabb7e118a3b4e019b62a6e48bR525
after we save the logs so users and use can get a chance to see those logs on the web

handlers/whatsapp/whatsapp.go
```diff
@@ -473,6 +475,12 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 				if attachmentCount == 0 {
 					mediaPayload.Caption = msg.Text()
 				}
+				mediaPayload.Filename, err = utils.BasePathForURL(mediaURL)
+
+				// Break out on error
```
norkans7, line 480: Adjust the comment

norkans7:
@rafaelbreno check the tests are failing

norkans7 (approved):

Issue #360, opened 2021-05-28
WhatsApp "interactive message" support


rudigiesler:
The WhatsApp API have added what they call "interactive messages". These are messages that allow users to more easily select an item from a list, with either up to 3 buttons, or a list with up to 10 items.

I think this will fit quite nicely into the "Quick Reply" functionality in RapidPro flows, but there might need to be some hard-coded logic here if we don't want to change the flow editor (eg. 1-3 items = buttons, 3-10 items = list)

I haven't looked into, or tried this functionality yet, but I wanted to get the ball/discussion rolling here so long, to decide if this is something that RapidPro wants to support, and if so, what is the best way to implement it.

Documentation: https://developers.facebook.com/docs/whatsapp/api/messages#interactive-object

matmsa27:
This feature is a good improvement from WhatsApp. The quick reply is a feature long-awaited from our Rapidpro customers. To implement this on Rapidpro, I want to raise two points, that I want to discuss with you.

1 - This feature for interactive messages, is only supported in a version from v2.35.2 WhatsApp API. For this, to use this feature is necessary to have a WhatsApp Docker instance with this version on minimum.
2 - The Whatsapp handler in courier is used to other BSPs, like 360, Textit, and to docker instance. For this, if using a channel like this and the API from WhatsApp doesn't have on the minimum version (v2.35.2), this feature didn't work. And, it doesn't have the minimum version, how courier will treat this? Maybe a solution is to have the version from API on the channel to treat this in a better way when creating flows on Rapidpro to support this? Just ideas and points that I have been thinking about and want to discuss with you.

rudigiesler:
@matmsa27 There have been other features that are only supported on newer versions of the WhatsApp API before, I wonder how those were handled? eg. we started with "highly structured messages" / hsm, then the API moved over to "template messages", which had a different structure.

I'm not 100% sure, but I know that there's a config field on the channel whether to use the hsm/template format in the API.

nicpottier:
Ya, I think adding this definitely makes sense, agree that buttons for quick replies 1-3 and list for more than that.

I guess we'd have to test it out but I bet because `interactive` is a new node in JSON it is just ignored by versions of WhatsApp that don't support it, generally extra JSON works that way. If so, that would make it easy to add for the time being.

If either one of you want to lead adding this, it should be a pretty easy change on the Courier side.

johncordeiro:
hey @nicpottier , I've tried to send in older versions and that's the result:
```
{
[...]

PR #361, opened 2021-06-02
Add handling for whatsapp interactive message throught quick reply

rasoro:
This allow Courier to handle quickreply message from RapidPro and send as interactive message. And also allow Rapidpro to handle interactive message responses by interpreting the button text or the option title selected in list as message text.

More details: #360


handlers/whatsapp/whatsapp.go
```diff
@@ -593,25 +658,106 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 
-				// add logs to our status
-				for _, log := range logs {
-					status.AddLog(log)
+			qrs := msg.QuickReplies()
+
+			channel := msg.Channel()
+
+			wppVersion := channel.ConfigForKey("version", "0").(string)
+			hasCompatibleVersion := wppVersion >= interactiveMsgMinSupVersion
```
nicpottier, line 667: This is not going to work for string versions. You'll need to use something like the semantic versioning package `semver`

nicpottier: Also we need the RapidPro side of this as well right?

handlers/whatsapp/whatsapp.go
```diff
@@ -593,25 +658,106 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+						status.AddLog(log)
+					}
+					if err != nil {
+						break
+					}
+
+					// if this is our first message, record the external id
+					if i == 0 {
+						status.SetExternalID(externalID)
+					}
```
nicpottier, line 760: A lot of repetition above, can you refactor this so we aren't repeating `sendWhatsAppMsg`, logging and external id over and over? Maybe it's a callback function that's passed in to build the payload?

nicpottier (changes requested):

handlers/whatsapp/whatsapp.go
rasoro: Hello, nicpottier. To improve version checking i used a semver package in the new commit as you recommended. Does it look better that way?

handlers/whatsapp/whatsapp.go
rasoro: I think your suggestion makes a lot of sense. So I tried to improve it that way, what did you think of how it is now? any suggestions are welcome and i will always be willing to try to improve.

handlers/whatsapp/whatsapp.go
```diff
@@ -33,6 +34,8 @@ const (
 	channelTypeTXW = "TXW"
 
 	mediaCacheKeyPattern = "whatsapp_media_%s"
+
+	interactiveMsgMinSupVersion = "v2.35.2"
```
nicpottier, line 38: might as well spell out `Supported` here.

nicpottier (approved):
Definitely looks better, thanks! Is there a rapidpro side of this yet to populate the version?

norkans7:
I think on this we need to have the channel work right away once we add it.
Currently the channel will be broken until the version is synced with the templates.

I think the interactive messages can be the default behavior for WA channel type but have a migration on RapidPro to set a flag(my preference) or the version to the current channels and then new channels without that set are considered compatible

matmsa27:
> I think on this we need to have the channel work right away once we add it.
> Currently the channel will be broken until the version is synced with the templates.

Yep, I agree with you @norkans7. To merge this is necessary the [task for template message](https://github.com/nyaruka/rapidpro/pull/3364) get the version from the WhatsApp instance and set the version on channel config for each channel.

> I think the interactive messages can be the default behavior for WA channel type but have a migration on RapidPro to set a flag(my preference) or the version to the current channels and then new channels without that set are considered compatible

For each channel created, the task for sync templates will do the job to set the version for each channel. To new channels, what do you think do an adjustment on creating a channel, to get the version on WhatsApp instance and set key `version` on channel config?

norkans7:
I think as this feature will be supported going forward, it is best the channel that is added without the version set to work with those feature enabled.

And a migration on RapidPro to set the version on the existing channels to set the version so we skip the unsupported features. So the use of the flag of version is temporal as we migrate the channels to the latest version

the task to sync the version will take a while to populate the version as of https://github.com/nyaruka/rapidpro/pull/3364

nicpottier:
@rasoro If you can fix the conflicts we're ready to merge this now that the RapidPro side is done.

rasoro:
hi, @nicpottier . Thanks for letting me know, fixed.

nicpottier:
Thanks!

PR #363, opened 2021-07-19
🚀 Latest gocommon

rowanseymour:


.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 on: [push, pull_request]
 env:
-  go-version: '1.15.x'
+  go-version: '1.16.x'
```
rowanseymour, line 4: needed because gocommon includes our locales library which uses `embed` to include the locale data

PR #364, opened 2021-07-21
✈️ Always set `sent_on` for W/S/D statuses if not already set

rowanseymour:


backends/rapidpro/backend_test.go
```diff
@@ -480,16 +483,26 @@ func (ts *BackendTestSuite) TestMsgStatus() {
 	m, err := readMsgFromDB(ts.b, courier.NewMsgID(10001))
 	ts.NoError(err)
 	ts.Equal(m.Status_, courier.MsgWired)
-	ts.Equal(m.ExternalID_, null.String("ext0"))
 	ts.True(m.ModifiedOn_.After(now))
 	ts.True(m.SentOn_.After(now))
 
-	// update by id, no external id, shouldn't overwrite it
```
rowanseymour, line 487: some of these comments don't match what the test is doing..

nicpottier (approved):

backends/rapidpro/status.go
```diff
@@ -265,11 +265,11 @@ UPDATE msgs_msg SET
 		END,
 	sent_on = CASE 
 		WHEN 
-			s.status = 'W' 
+			s.status IN ('W', 'S', 'D') AND sent_on IS NULL
```
norkans7, line 268: what about the msgs with R? we clear the sent on on them?

rowanseymour: we don't use `R` any more - resending requeues messages as QUEUED

norkans7: never mind we use pending on them

norkans7 (approved):

PR #365, opened 2021-07-21
Make DBMsg.SentOn nullable

rowanseymour:


backends/rapidpro/backend_test.go
```diff
@@ -480,23 +480,21 @@ func (ts *BackendTestSuite) TestMsgStatus() {
 	ts.NoError(err)
 	time.Sleep(time.Second)
 
-	m, err := readMsgFromDB(ts.b, courier.NewMsgID(10001))
-	ts.NoError(err)
+	m := readMsgFromDB(ts.b, courier.NewMsgID(10001))
```
rowanseymour, line 483: this func is just for testing and I noticed a place where we weren't checking the err result.. which was actually an error because it couldn't scan sent_on=NULL into a `time.Time`... so now it just panics

norkans7 (approved):

PR #367, opened 2021-07-21
Give S3 storage test new context

rowanseymour:
Update to latest gocommon adds context param to Storage.Test(..)

Noticed pushing to staging we got an error saying "S3 storage not available" but it was working fine https://sentry.io/organizations/nyaruka/issues/2525368566/?project=181813&query=is%3Aunresolved

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -700,7 +700,9 @@ func (b *backend) Start() error {
 	}
 
 	// test our storage
+	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
 	err = b.storage.Test(ctx)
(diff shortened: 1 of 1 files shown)


backends/rapidpro/backend.go
```diff
@@ -700,7 +700,9 @@ func (b *backend) Start() error {
 	}
 
 	// test our storage
+	ctx, cancel = context.WithTimeout(context.Background(), 3*time.Second)
 	err = b.storage.Test(ctx)
```
rowanseymour, line 704: this re-using the same context from above which has already been canceled

norkans7 (approved):

Salah856:
great

PR #372, opened 2021-08-12
Weni Web Chat Handler

paulobernardoaf:
https://github.com/rapidpro/rapidpro/issues/1429


handlers/weniwebchat/weniwebchat.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, handlers.WriteAndLogRequestIgnored(ctx, h, channel, w, r, "ignoring request, unknown message type")
+	}
+
+	// check empty content
+	if payload.Message.Text == "" && payload.Message.MediaURL == "" && (payload.Message.Latitude == "" || payload.Message.Longitude == "") {
+		return nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r, errors.New("blank message, media or location"))
+	}
+
+	// build urn
+	urn, err := urns.NewURNFromParts(urns.ExternalScheme, payload.From, "", "")
```
rowanseymour, line 73: Any reason to not define a new URN type? And what is `From` here?

handlers/weniwebchat/weniwebchat.go
```diff
@@ -0,0 +1,259 @@
...
+		Type: payType,
+		To:   to,
+		From: from,
+		Message: moMessage{
+			QuickReplies: quickReplies,
+		},
+	}
+}
+
+func getTimestamp() string {
```
norkans7, line 253: why not use a local variable where the timestamp is needed?

handlers/weniwebchat/weniwebchat.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	for _, log := range logs {
+		status.AddLog(log)
+	}
+
+	return status, nil
+}
+
+func newOutgoingMessage(payType, to, from string, quickReplies []string) *moPayload {
```
norkans7, line 242: I think this method is not needed as it is called just once

handlers/weniwebchat/weniwebchat.go
```diff
@@ -0,0 +1,259 @@
...
+	MediaURL     string   `json:"media_url,omitempty"`
+	Caption      string   `json:"caption,omitempty"`
+	Latitude     string   `json:"latitude,omitempty"`
+	Longitude    string   `json:"longitude,omitempty"`
+	QuickReplies []string `json:"quick_replies,omitempty"`
+}
+
+func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStatus, error) {
+	start := time.Now()
+	status := h.Backend().NewMsgStatusForID(msg.Channel(), msg.ID(), courier.MsgSent)
```
norkans7, line 126: The status should start assuming it is an error until we are sure the message was sent.

handlers/weniwebchat/weniwebchat.go
```diff
@@ -0,0 +1,259 @@
...
+			} else if strings.HasPrefix(mimeType, "video") {
+				payload.Message = moMessage{
+					Type:     "video",
+					MediaURL: attachmentURL,
+				}
+			} else {
+				elapsed := time.Since(start)
+				log := courier.NewChannelLogFromError("Error sending message", msg.Channel(), msg.ID(), elapsed, fmt.Errorf("unknown attachment mime type: %s", mimeType))
+				logs = append(logs, log)
+				status.SetStatus(courier.MsgFailed)
```
norkans7, line 170: we usually mark a message as errored so we can retry that up to 3 times, 
And in this case I think you want to just log the error as other parts could have been sent.

norkans7 (changes requested):
Check the comments and match the URN scheme on RapidPro https://github.com/nyaruka/rapidpro/pull/3488/files#diff-7b90ea5e3fb9ba05a8e735870b6aca87d34d2c41584165c9717416ebdd0a12e2R73

handlers/weniwebchat/weniwebchat.go
rasoro: waiting for nyaruka/gocommon#51 be approved to proceed

matmsa27:
Hey guys, any news on this review?

rowanseymour:
@matmsa27 hey I still have some queries about use of URNs - so we've agreed on a URN format like `webchat:<user>@<site>` but where are you going to get `<user>` and `<site>` from?

paulobernardoaf:
Hey @rowanseymour , both `user` and `site` are fields that can be configured on the Web Script, but they are not mandatory.
If not provided, the Web Script will generate a random `user` hash, and get the current window domain to use it as `site`.

When the Web Script is loaded, those fields are computed and then sent to the WebSocket to check if it is a new session or an existing one. After that, the WebSocket will communicate with Rapidpro using that urn scheme as an identifier.

rowanseymour:
@paulobernardoaf my concern is that if those are things that the client just sets - then what's to stop me calling myself Paulo and hijacking your existing contact in the RapidPro instance?

rasoro:
> @paulobernardoaf my concern is that if those are things that the client just sets - then what's to stop me calling myself Paulo and hijacking your existing contact in the RapidPro instance?

for that, you would first need to authenticate yourself on the site with paulo's credentials to get his auth||identifier||data and assemble his user based on that auth. example urn: <paulo:auth||identifier||data>@<site>.

for sites that allow anonymous contacts the <user> is a random hash, as Paulo mentioned.

rasoro:
any other questions, @rowanseymour ?

ericnewcomer:
Hey guys I do think maybe we aren't quite on the same page on the goals for this, and perhaps a call is in order.

For me it is important that by adding a web chat channel that it is a first class multi tenant feature within RapidPro. ie,

  * User adds new Web Chat channel
  * Channel settings expose the embed code for the widget (hosted on rapid)
  * Admins can configure the web chat widget on the channel page (ie, apply their own colors or branding)

All very similar to how other real time chat services work.

This implies that the supporting infrastructure, both the client and the server for the webchat should be a native part of RapidPro and not an external dependency nor branded as a different thing.

matmsa27:
Hey, @ericnewcomer thanks for your contribution here. I disagree with you, the Weni Web Channel, in the beginning, we believed that this chat could be inside the Rapidpro, for example, the socket could be inside the courier and the other pieces on the Rapidpro, but as we discussed [rapidpro/rapidpro#1429](https://github.com/rapidpro/rapidpro/issues/1429) we decided to implement these things separated following the premisses to be an open-source application like that courier, mailroom, indexer and archiver are microservices implemented in go lang that can use in an independent way, the Weni web chat application can be used in an independent way too but is an extensible application. 

This application sends and receives messages integrated with other platforms like Rapidpro. Weni web chat is a channel communication like Telegram, Facebook, WhatsApp and etc that can have integration with Rapidpro.

ericnewcomer:
Cleary our goals are not aligned. I guess my point is that a first class Web Chat channel has far greater value to the community than just another external channel type. Additionally, RapidPro itself is migrating towards a chat-first interface in places. To me, not trying to address this more generically is a mistake.

I'm sure you can appreciate that Weni Web Chat is not truly like Telegram, Facebook, WhatsApp, etc. These channels have billions of active users. Is there any reason you don't just want to maintain this particular channel type on your fork? We won't be running it and thusly can't really be on the hook to maintain or support it anyways, so not sure there's value to pulling it in.

matmsa27:
Hi @ericnewcomer , thanks for sharing our discussion according made on the call. This month we are working on these improvements and we will update these PRs.

PR #375, opened 2021-08-23
Fix to only set the quick replies keyboard for the last message

norkans7:


handlers/telegram/telegram.go
```diff
@@ -188,29 +188,33 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
 		}
 
-		externalID, log, err := h.sendMsgPart(msg, authToken, "sendMessage", form, keyboard)
+		externalID, log, err := h.sendMsgPart(msg, authToken, "sendMessage", form, msgKeyBoard)
 		status.SetExternalID(externalID)
 		hasError = err != nil
 		status.AddLog(log)
 
-		// clear our keyboard which has now been sent
-		keyboard = nil
```
rowanseymour, line 209: I'm not getting why this doesn't work as is..

norkans7: We are setting the keyboard on the first message instead of the last one so the last one is having `"remove_keyboard":true`

norkans7: The change is to make sure we always set the keyboard only on the last one

rowanseymour: Ah ok - it needs to be on the last rather than the first! That make sense. 

Maybe cleaner to still create one keyboard instance at the top if `len(qrs) > 0`

norkans7: Done

rowanseymour (approved):

Issue #366, opened 2021-07-21
WhatsApp Template Message Media Support


matmsa27:
Hello everyone, WhatsApp provides support to template message with media, but on Rapidpro only support Template Message with Text. Can we planning here to support this?

Reference: https://developers.facebook.com/docs/whatsapp/api/messages/message-templates/media-message-templates/

rasoro:
I believe this would be possible by adding new fields to the templatePayload struct that already exists.

Highlighting here some things we noticed in small tests we did. As an example media files receive url where the files are stored, and with that we would take advantage of the attachments tab in the "send message".

but we would have to see in more detail how to handle the template update tasks.

matmsa27:
Hi @nicpottier and @rowanseymour, what do you think about this on Rapidpro?

matmsa27:
Hello everyone, can we discuss this point? we are very excited here to implement this and open the PR.

matmsa27:
Hi @ericnewcomer , I would like to discuss this with you, what do you think?

rowanseymour:
I think we can certainly add support for this and I'm trying to understand what that entails. Does this require supporting a new API version? I'm looking at the existing template sync code and it assumes templates are made up of component types `HEADER`, `BODY`, `FOOTER`, but these docs show templates as being `header|image|document`, `body`..

rasoro:
I would like to mention that, i think the component types still are `HEADER`, `BODY` and `FOOTER`, but only in `HEADER` component has a field named `FORMAT` that can be: `TEXT` || `IMAGE` || `DOCUMENT` || `VIDEO`.

https://developers.facebook.com/docs/whatsapp/business-management-api/message-templates#creating-message-templates

Robi9:
For the addition of a media message template, the following modifications were thought:
To find out if it is a media or text template, an identification field must be added so that it is possible to know if that object created in `RapidPro` is a template and what its type is (media or text).

If it is identified that the template is a media template, in the `Courier` the modification of the `TemplatePayload` must occur, in which the `Components` can be of 3 types: `header`, `body` and `footer`. These types must be added, currently only the `body` type is implemented. In addition, `Param` must also accept `text`, `currency`, `date_time`, `image`, `document` and `video`. Currently `Param` only accepts `text` and it already fulfills the role of `currency` and `date_time` types, so would it still be necessary to add these types?

[...]

PR #378, opened 2021-09-09
Supporting media message template from whatsapp.

Robi9:
Openning this PR to support template media message from whatsapp according to [this issue](https://github.com/nyaruka/courier/issues/366).


matmsa27:
Hi @rowanseymour , any news to review this PR? We running this on our fork for the last 3 weeks and works properly.

handlers/whatsapp/whatsapp.go
```diff
@@ -564,76 +582,11 @@ func buildPayloads(msg courier.Msg, h *handler) ([]interface{}, []*courier.Chann
...
-					To:   msg.URN().Path(),
-					Type: "document",
-				}
-				if attachmentCount == 0 {
-					mediaPayload.Caption = msg.Text()
-				}
-				mediaPayload.Filename, err = utils.BasePathForURL(mediaURL)
+	// do we have a template?
+	var templating *MsgTemplating
+	templating,err = h.getTemplate(msg)
```
rowanseymour, line 587: looks like gofmt hasn't run? And this can be `templating, err := `

rowanseymour:
@matmsa27 sorry this fell off the radar - the diff looks like `gofmt` hasn't been run as it's kinda hard to read - can someone clean that up

matmsa27:
Hi @rowanseymour , no problem. Sure, thanks for the feedback and support. @Robi9 will run and update the PR.

Robi9:
I made the changes that were suggested.

rowanseymour (approved):

norkans7:
I am wondering how we are syncing the image templates on RapidPro side as we only consider text templates @Robi9 @matmsa27 

https://github.com/nyaruka/rapidpro/blob/e6b69b1f76963d19ad7b6d145f6423a9583f6d87/temba/utils/whatsapp/tasks.py#L116

What happens if we try to add an attachment on a template that did not have the image defined?
How do we know the templates where the attachment was defined or not?

matmsa27:
Hi @norkans7 , actually we are only supporting sending the template message with this PR on Courier side on this phase. I believe that we need more improvements on the Rapidpro Django side to sync the template messages with media and on the flow editor to support the template with media to be more clear to the user about the template selected to send on the card. 

On the next steps is interesting to improve the models about the templates, sync method template to save information about the type of template (media or text), if media save what the type of media (document, video or image). And in another side change the flow editor to knows when template media is selected to guide the user add the correct media to template.

About the syncing, we are putting the templates manually on the RapidPro database, and with this PR we select the template on flow editor and add the attachment according to the template, this works as well in this first stage.

> What happens if we try to add an attachment on a template that did not have the image defined?
The courier will try send the template with media, and the wpp will return an error. The idea is, the user that is configuring the flow and the template message knows if the template support template message or not.

> How do we know the templates where the attachment was defined or not?
Actually, only the users that have access on whatsapp account on facebook business and the organization on Rapidpro.

matmsa27:
I believe this PR is a first phase to support this in general on Rapidpro. 

> I think this can wait the changes you mentioned needed on RapidPro and the flow editor and the use of attachments seems like a quick hack but will be broken when we add the proper support for the media of templates as the editor will probably have a field for the media as we do for the variables..

Can we do a migration to change the media setuped to the new way when flow editor will change?

We are open to implement this on Rapidpro django side and discuss the changes on flow editor. We can prototype and discuss before the implementation starts.

PR #382, opened 2021-10-08
Throttle courier queues when the channel has rate limit redis key not…

norkans7:
… expired


nicpottier:
This feels slightly odd in that it seems like we are basically using a redis key as a communication channel to move a channel into being throttled when we next try to pop a message off. Why are we doing it that way as opposed to moving the channel to the throttled list explicitly when we decide we are throttled? I guess the advantage of this method is we can set how long we throttle the channel for by tweaking the expiration but the downside is that every call to pop a message off a queue now has an extra lookup. I think I'd rather see our throttled logic expanded to just allow throttles more than a second if that's what is needed:
  https://github.com/nyaruka/courier/blob/main/queue/queue.go#L212
  
That could work by having both the queue and say a counter in the throttle list and that gets decremented in the dethrottler, moving anything that now has a count <= 0 to the active queue set.

handlers/whatsapp/whatsapp.go
```diff
@@ -839,6 +843,16 @@ func sendWhatsAppMsg(msg courier.Msg, sendPath *url.URL, payload interface{}) (s
 	req, _ := http.NewRequest(http.MethodPost, sendPath.String(), bytes.NewReader(jsonBody))
 	req.Header = buildWhatsAppHeaders(msg.Channel())
 	rr, err := utils.MakeHTTPRequest(req)
+
+	if rr.StatusCode == 429 || rr.StatusCode == 503 {
+		rateLimitKey := fmt.Sprintf("rate_limit:%s", msg.Channel().UUID().String())
+		rc.Do("set", rateLimitKey, "engaged")
+		rc.Do("expire", rateLimitKey, 2)
```
rowanseymour, line 850: doesn't the 429 response tell us when we can retry ?

norkans7: No, we get 

```
{"meta":{"api_status":"stable","version":"2.35.4"},"errors":[{"code":1015,"title":"Too many requests","details":"Rate limiting engaged - Too Many Requests."}]}
```

rowanseymour: so we're backing down for 2 seconds but do we have any information if that's a reasonable back down?

norkans7: We are pausing the queue to send for 2 seconds and try again once we have the 429 status we pause for another 2 seconds

norkans7: https://developers.facebook.com/docs/whatsapp/api/rate-limits#capacity

The rate limit we are having is that we had more than 50 requests per second so stopping 2 seconds seems that will help reset the limit

rowanseymour: ok can you add a little comment to the code here explaining the rational for 2 and maybe mention that in future it should use header values

rowanseymour: and then let's get this out there!

rowanseymour (approved):

PR #387, opened 2021-11-10
Adds the Africa's Talking 'Expired' Status in Its Handler

jasonrogena:
Add support for the 'Expired' status from Africa's Talking which seems
to be thrown but isn't handled. Courier currently throws the following
error when it encounters this status:

time="2021-11-10T10:29:10+03:00" level=info msg="request errored"
channel_uuid=<UUID> elapsed_ms=0.109525 error="unknown status
'Expired', must be one of 'Success','Sent', 'Buffered','Rejected'
or 'Failed'" url=/c/at/<UUID>/status

Mark messages that have this status as failed as it doesn't appear
messages that have this status get delivered to users.

Signed-off-by: Jason Rogena <jason@rogena.me>


rowanseymour (approved):
we're in code freeze until the end of the week, but looks good!

PR #392, opened 2021-12-17
Instagram handler

matmsa27:
According to [issue](https://github.com/rapidpro/rapidpro/issues/1516) and PR opened on [Rapidpro repository](https://github.com/rapidpro/rapidpro/pull/1596).


norkans7:
@matmsa27 I see most codes are copied from the Facebook app channel type?
What challenges are there to use the same app for Facebook and Instagram? 
So the app will need to have all the required permissions for both?

Could that simplify the Instagram handler and possible make a few adjustment to the facebook app one to accommodate that as well?

Robi9:
> What challenges are there to using the same app for Facebook and Instagram?

Yes, most of the code is very similar to Facebook. I don't see that it would have a problem to use the same application, it would just leave us stuck in having to use the same for both, we can have rapidpro hosts that are only interested in one of the channels so it would be interesting to have this application distinction.

> So the app will need to have all the required permissions for both?

In case it is just one app for both types, it would be necessary to have all permissions for Facebook and Instagram, but if it is just an app for Instagram, it would not need to have all permissions, it would only be some that the Facebook channel uses and Instagram permissions.

> Could that simplify the Instagram handler and possibly make a few adjustments to the Facebook app one to accommodate that as well?

Could you explain better what you thought about simplifying the Instagram handler if the same app is used for both?

norkans7:
@Robi9 I think an approach as the way we handle Twilio types that share a lot of code
https://github.com/nyaruka/courier/blob/main/handlers/twiml/twiml.go#L58

so the Facebookapp type will be the one with the handler but we can initiate the IG channel type are well in the file

all incoming messages will be receive on `/c/fba/receive` and we switch the channel we look for using the payload object value in the request we receive

That way I think we remain with less code so easier to maintain

Robi9:
@norkans7 Understood, I will adapt the Facebookapp handler to support the IG type.

Robi9:
@norkans7 I decided to separate receiving messages, so Facebook's keep arriving at `c/fba/receive` and Instagram's arrive at `c/ig/receive`, we can do that since the application allows us to indicate callback endpoints for types of different payload objects.

matmsa27:
Hi, @norkans7 We are already ready to receive the feedbacks about this PR.

norkans7:
@Robi9 good to have two webhooks configured, now I think we need to have the signature validation and webhook validation work for both.
And what is the use of `InstagramApplicationSecret` I think the app will be the same as the facebook one, correct?

handlers/facebookapp/facebookapp_test.go
```diff
@@ -612,7 +936,71 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
+	{Label: "Error",
+		Text: "Error", URN: "facebook:12345",
+		Status:       "E",
+		ResponseBody: `{ "is_error": true }`, ResponseStatus: 403,
+		SendPrep: setSendURL},
+}
+
+var SendTestCasesIG = []ChannelSendTestCase{
+	{Label: "Plain Send",
+		Text: "Simple Message", URN: "facebook:12345",
```
norkans7, line 955: These tests should us `instagram` URNs

norkans7: not `facebook` URNs

norkans7 (changes requested):
Adjust the instagram tests

handlers/facebookapp/facebookapp.go
```diff
@@ -56,18 +56,20 @@ const (
 	payloadKey    = "payload"
 )
 
+func newHandler(channelType courier.ChannelType, name string, validateSignatures bool) courier.ChannelHandler {
```
norkans7, line 59: `validateSignatures` is misleading as that parameter is for using channel UUIDS routes or not

handlers/facebookapp/facebookapp_test.go
```diff
@@ -502,50 +785,82 @@ func buildMockFBGraph(testCases []ChannelHandleTestCase) *httptest.Server {
...
+				metadata, _ := handler.DescribeURN(context.Background(), testChannels[0], tc.urn)
+				assert.Equal(t, metadata, tc.metadata)
+			}
+		} else {
+			handler := newHandler("IG", "Instagram", false).(courier.URNDescriber)
+			tcs := []struct {
+				urn      urns.URN
+				metadata map[string]string
+			}{
+				{"facebook:1337", map[string]string{"name": "John Doe"}},
```
norkans7, line 829: `instagram` URN

handlers/facebookapp/facebookapp_test.go
```diff
@@ -502,50 +785,82 @@ func buildMockFBGraph(testCases []ChannelHandleTestCase) *httptest.Server {
...
 		// no name
-		w.Write([]byte(`{ "first_name": "", "last_name": ""}`))
+		w.Write([]byte(`{ "name": ""}`))
 	}))
 	graphURL = server.URL
 
 	return server
 }
 
 func TestDescribe(t *testing.T) {
```
norkans7, line 800: I think it is best to have 2 separate tests one for FBA and the other for IG instead of the for loop on a slice of slices and the if condition

norkans7 (approved):

handlers/facebookapp/facebookapp.go
```diff
@@ -172,19 +175,24 @@ func (h *handler) GetChannel(ctx context.Context, r *http.Request) (courier.Chan
...
+		return nil, fmt.Errorf("object expected 'page' or 'instagram', found %s", payload.Object)
 	}
 
 	// no entries? ignore this request
 	if len(payload.Entry) == 0 {
 		return nil, fmt.Errorf("no entries found")
 	}
 
-	pageID := payload.Entry[0].ID
+	EntryID := payload.Entry[0].ID
```
rowanseymour, line 188: `entryID`

handlers/facebookapp/facebookapp_test.go
```diff
@@ -387,25 +578,58 @@ var notPage = `{
...
+
+var noMessagingEntriesIG = `{
+	"object":"instagram",
+	"entry": [{
+		"id": "12345"
 	}]
 }`
 
-var unkownMessagingEntry = `{
+var unkownMessagingEntryFBA = `{
```
rowanseymour, line 610: unknown.. not your typo but you can fix while you're here

rowanseymour (changes requested):

rowanseymour (approved):

PR #394, opened 2021-12-30
Quick Replies for VK

Robi9:
This PR is about supporting quick replies to VK handler. VK treats quick replies as keyboard, [here](https://vk.com/dev.php?f=4.%20Bot%20keyboards&method=bots_docs_3) we can see in more detail about keyboards.

### DIFF
--- a/handlers/vk/vk.go
+++ b/handlers/vk/vk.go
@@ -56,6 +56,7 @@ var (
 	paramMessage      = "message"
 	paramAttachments  = "attachment"
 	paramRandomId     = "random_id"
+	paramKeyboard     = "keyboard"
 
 	// base upload media values
 	paramServerId = "server"
@@ -113,6 +114,7 @@ type moNewMessagePayload struct {
 					Lng float64 `json:"longitude"`
 				} `json:"coordinates"`
 			} `json:"geo"`
+			Payload string `json:"payload"`
 		} `json:"message" validate:"required"`
 	} `json:"object" validate:"required"`
 }
@@ -187,6 +189,23 @@ type photoUploadPayload struct {
 type mediaUploadInfoPayload struct {
 	MediaId int64 `json:"id"`
 	OwnerId int64 `json:"owner_id"`
+}
+
+type Keyboard struct {
+	One_Time bool              `json:"one_time"`
+	Buttons  [][]buttonPayload `json:"buttons"`
+	Inline   bool              `json:"inline"`
+}
+
+type buttonPayload struct {
+	Action buttonAction `json:"action"`
+	Color  string       `json:"color"`
+}
+
+type buttonAction struct {
+	Type    string `json:"type"`
+	Label   string `json:"label"`
+	Payload string `json:"payload"`
 }
 
 // receiveEvent handles request event type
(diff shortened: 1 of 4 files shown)


handlers/vk/vk_test.go
```diff
@@ -445,6 +472,25 @@ var sendTestCases = []ChannelSendTestCase{
 			},
 		},
 	},
+	{
+		Label:        "Send keyboard",
+		Text:         "Send keyboard",
+		URN:          "vk:123456789",
+		QuickReplies: []string{"Yes"},
```
norkans7, line 479: Can you make the test with many quick replies?

norkans7:
This will add all  buttons in one row, I think we need to probably make a new row every 3 buttons or better use the length of the button text to know when to make a new row

Robi9:
I did some tests and regardless of the text size, in a new line, 5 buttons are accepted. When the text on these buttons is large, they end up showing a small portion of that text and hiding the rest.

handlers/vk/vk.go
```diff
@@ -417,6 +445,58 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+		keyboard.Buttons = append(keyboard.Buttons, row)
+
+	}
+	keyboard.One_Time = true
+
+	return keyboard
+}
+
+func AddTextButton(label string, payload interface{}, color string, keyboard Keyboard) Keyboard {
+	b, err := json.Marshal(payload)
```
rowanseymour, line 480: I'm fine with not error checking JSON marshalling of things we control - but use `jsonx.MustMarshal`

handlers/vk/vk.go
```diff
@@ -417,6 +445,58 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 	return status, nil
 }
 
+// ToJSON returns the JSON encoding of Keyboard.
+func ToJSON(keyboard Keyboard) string {
+	b, _ := json.Marshal(keyboard)
+	return string(b)
+}
```
rowanseymour, line 452: this function doesn't feel necessary - use `jsonx.MustMarshal`

handlers/vk/vk.go
```diff
@@ -417,6 +445,58 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
...
+
+//keyboard
+func buildQuickRepliesParams(msg courier.Msg) Keyboard {
+
+	var keyboard Keyboard
+	keyboard = AddRowKeyboard(keyboard)
+
+	for _, qr := range msg.QuickReplies() {
+		keyboard = AddTextButton(qr, qr, "secondary", keyboard)
+	}
```
rowanseymour, line 462: maybe we can do something similar to the telegram handler and use `utils.StringsToRows` to create better layouts when things don't fit on one row?

rowanseymour (changes requested):

Robi9:
@rowanseymour I sought the same approach taken for the telegram, it applied well to the VK.

norkans7 (approved):

rowanseymour (approved):

PR #400, opened 2022-01-18
Update HX user_data parameter

norkans7:


rowanseymour:
@norkans7 @ericnewcomer maybe before fixing this, we start queuing messages in mailroom with their flow name and use that?

norkans7:
Sure, I had just assumed that was broken because the parameter was changed(or we had that wrong all the time)

rowanseymour:
I can make the change on the mailroom side but are we good with flow name? UUID? I'm leaning toward name..

norkans7:
I think mailroom sends both the same as the archiver so we are very specific and HX we can only send one of those if that is what is needed

ericnewcomer (changes requested):
We actually want to change the `user_data` param to be the flow name

backends/rapidpro/msg.go
```diff
@@ -518,6 +518,8 @@ type DBMsg struct {
 	SessionWaitStartedOn_ *time.Time `json:"session_wait_started_on,omitempty"`
 	SessionStatus_        string     `json:"session_status,omitempty"`
 
+	Flow_ json.RawMessage `json:"flow,omitempty"`
```
rowanseymour, line 521: no reason to not parse it all in one go.. you can even use a anon struct like 

```
Flow_ *struct {
    UUID string `json:"uuid"`
    Name string `json:"name"`
} `json:"flow"`
``

**Review by rowanseymour — CHANGES_REQUESTED** [review]

File: backends/rapidpro/msg.go (hunk 1 above)
Comment (norkans7) [reply backends/rapidpro/msg.go:521]: What type will `Flow()` will return there?

Comment (norkans7) [reply backends/rapidpro/msg.go:521]: never mind no need for anymore

Comment (rowanseymour) [reply backends/rapidpro/msg.go:521]: if you need a type just add something like `MsgFlowRef`

Comment (norkans7) [reply backends/rapidpro/msg.go:521]: I am confused on how to use that I will try that again tomorrow

File: handlers/highconnection/highconnection_test.go [hunk 2]
```diff
@@ -57,14 +57,35 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
+		URN:    "tel:+250788383383",
+		Status: "W",
+		URLParams: map[string]string{
+			"accountid":  "Username",
+			"password":   "Password",
+			"text":       "Simple Message",
+			"to":         "+250788383383",
+			"ret_id":     "10",
+			"datacoding": "8",
+			"user_data":  "",
```
norkans7, line 88: A message not from a flow will have the user_data as empty string

msg.go
```diff
@@ -78,6 +78,13 @@ func NewMsgUUIDFromString(uuidString string) MsgUUID {
 	return MsgUUID{uuid}
 }
 
+type MsgFlowRef struct {
+	UUID string `json:"uuid" validate:"uuid4"`
+	Name string `json:"name"`
+}
+
+var NilMsgFlowRef = MsgFlowRef{"", ""}
```
rowanseymour, line 86: We don't want to serialize messages with `"flow": {"uuid": "", "name": ""}`.. should be a pointer on `Msg` which can be nil if there's no flow.

While you're changing this let's call it just `FlowReference` for simplicity and consistency with mailroom/goflow.

rowanseymour (changes requested):

handlers/test.go
```diff
@@ -89,6 +89,10 @@ type ChannelSendTestCase struct {
 	HighPriority         bool
 	ResponseToExternalID string
 	Metadata             json.RawMessage
+	Flow                 *struct {
+		UUID string
+		Name string
+	}
```
rowanseymour, line 95: why not use the `FlowReference` type here?

handlers/highconnection/highconnection_test.go
```diff
@@ -57,14 +57,35 @@ var defaultSendTestCases = []ChannelSendTestCase{
 		Text:   "Simple Message",
 		URN:    "tel:+250788383383",
 		Status: "W",
+		Flow: &struct {
+			UUID string
+			Name string
+		}{UUID: "9de3663f-c5c5-4c92-9f45-ecbc09abcc85", Name: "Favorites"},
```
rowanseymour, line 63: use `FlowReference` type

rowanseymour (changes requested):

rowanseymour (approved):

PR #409, opened 2022-01-28
Set next_attempt in the next day for WA tiers limit hit

norkans7:


backends/rapidpro/status.go
```diff
@@ -103,7 +103,7 @@ UPDATE msgs_msg SET
 		WHEN 
 			:status = 'E' 
 		THEN 
-			NOW() + (5 * (error_count+1) * interval '1 minutes') 
+			GREATEST(NOW() + (5 * (error_count+1) * interval '1 minutes'), :next_attempt::timestamp with time zone)
```
rowanseymour, line 106: I'm not sure `:next_attempt::timestamp with time zone` is valid syntax - `:foo` syntax is just a way for the sqlx library to created queries that bind to struct fields but they ultimately get replaced with regular `$1`, `$2` etc param names.

norkans7:
replaced by #411

backends/rapidpro/status.go
norkans7: I changed to use a string for interval that we can handle well easily in the SQL

norkans7: #411

PR #411, opened 2022-02-02
Use interval to reschedule message next day for WA channel when they …

norkans7:
…hit the tiers limit


backends/rapidpro/status.go
```diff
@@ -102,9 +104,15 @@ UPDATE msgs_msg SET
...
-			NOW() + (5 * (error_count+1) * interval '1 minutes') 
-		ELSE 
+		THEN CASE
+			WHEN
+				:next_attempt_interval = '24 hours'
+			THEN
+				NOW() + interval '24 hours'
+			ELSE
+				NOW() + (5 * (error_count+1) * interval '1 minutes')
+			END
```
rowanseymour, line 114: Feels a bit odd that `next_attempt_interval` is only used if it's `'24 hours'` and otherwise ignored. We could pass the number of minutes and have it default to `5 * (error_count+1)`

```sql
NOW() + CASE WHEN :next_attempt_minutes > 0 THEN (:next_attempt_minutes || ' minutes')::interval ELSE (5 * (error_count+1) * interval '1 minutes') END;
```

rowanseymour:
Replaced by https://github.com/nyaruka/courier/pull/413

PR #414, opened 2022-02-03
Use UPPERCASE for redis commands

norkans7:


handlers/jiochat/jiochat.go
```diff
@@ -221,7 +221,7 @@ func (h *handler) getAccessToken(channel courier.Channel) (string, error) {
 	defer rc.Close()
 
 	cacheKey := fmt.Sprintf("jiochat_channel_access_token:%s", channel.UUID().String())
-	accessToken, err := redis.String(rc.Do(http.MethodGet, cacheKey))
+	accessToken, err := redis.String(rc.Do("GET", cacheKey))
```
norkans7, line 224: I think we should not use the variable there

Issue #383, opened 2021-10-15
misbehavior in the uploaded filename sent in whatsapp handler


rasoro:
despite this PR #356 , messages with uploaded files sent to whatsapp have a problem with the display name.
(but files from URLs show normal)

in this example I send two files, one .xlsx that was uploaded in rapidpro, and another pdf from a URL
![Captura de tela de 2021-10-15 09-04-59](https://user-images.githubusercontent.com/75167930/137485341-81e624b1-bdc4-4508-9dc9-fd77066804e7.png)

I identified this `if` sets mediaURL = "":
https://github.com/nyaruka/courier/blob/db924e11570fcf7f39d005b3be800a342e90a313/handlers/whatsapp/whatsapp.go#L578

causing the problem to happen on the line that sets the filename
https://github.com/nyaruka/courier/blob/db924e11570fcf7f39d005b3be800a342e90a313/handlers/whatsapp/whatsapp.go#L597

removing that `if`, it works fine
![Captura de tela de 2021-10-15 09-05-56](https://user-images.githubusercontent.com/75167930/137486573-6871a564-3560-4c46-b34c-eaf435b0d326.png)

however, when the file comes from uploaded it has this UUID in the name,
I would like to discuss how to improve send files in this regard, find an alternative to make the display more readable to the contact.

rasoro:
https://github.com/nyaruka/courier/pull/356#issuecomment-828599394 as nickpotier said in that comment, it seems to me to be a problem having the filename as the message text. (really is?)

so instead what do you think of setting the filename in s3 equal to the original name?

rasoro:
Hi, @norkans7 . Getting the url before the "if" block looks great, I tested it with a real phone number and it worked. Can we open a PR with this?

Additionally, we have received complaints from customers with the filename, displayed in attached messages, being a UUID instead of the original filename. The UUID may seem confusing to the contact. Could we open a PR on Rapidpro addressing this issue as well?

PR #419, opened 2022-02-15
fix whatsapp uploaded attachment file name

rasoro:
fixing bug with attachment filename in whatsap handler, as discussed here: https://github.com/nyaruka/courier/issues/383


norkans7 (approved):

handlers/whatsapp/whatsapp.go
```diff
@@ -605,7 +606,7 @@ func buildPayloads(msg courier.Msg, h *handler) ([]interface{}, []*courier.Chann
 				if attachmentCount == 0 && !isInteractiveMsg {
 					mediaPayload.Caption = msg.Text()
 				}
-				mediaPayload.Filename, err = utils.BasePathForURL(mediaURL)
```
burakturkoglu5454, line 608: 608

burakturkoglu5454 (commented):
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

Issue #418, opened 2022-02-14
Metadata fetch bug for new Facebook contacts


Robi9:
With the union of the Instagram handler with the Facebook handler, an update was carried out in the [search]( https://github.com/nyaruka/courier/blob/main/handlers/facebookapp/facebookapp.go#:~:text=func%20(h%20*handler)-,DescribeURN,-(ctx%20context.Context ) for the metadata for new contacts, which ended up influencing the appearance of a bug in the creation of these new contacts, this caused some Facebook contacts to be created without the username.

norkans7:
I think you mean this change https://github.com/nyaruka/courier/pull/392/files#diff-799bba3fe4dcf9897cf02b0de8b0701a579fdb1799382d109cc97a268662d1d6L653-L657

the method takes a channel in the arguments I think you can use the channel type to do the right thing for each.

Open a PR for that and remember tests

PR #423, opened 2022-02-18
Fix metadata fetching for new Facebook contacts

Robi9:
Fix for bug discussed in issue #418

### DIFF
--- a/handlers/facebookapp/facebookapp.go
+++ b/handlers/facebookapp/facebookapp.go
@@ -662,20 +662,34 @@ func (h *handler) DescribeURN(ctx context.Context, channel courier.Channel, urn 
 	base, _ := url.Parse(graphURL)
 	path, _ := url.Parse(fmt.Sprintf("/%s", urn.Path()))
 	u := base.ResolveReference(path)
-
 	query := url.Values{}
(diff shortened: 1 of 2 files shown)


handlers/facebookapp/facebookapp.go
```diff
@@ -662,20 +662,34 @@ func (h *handler) DescribeURN(ctx context.Context, channel courier.Channel, urn
...
+		query.Set("fields", "first_name,last_name")
+		query.Set("access_token", accessToken)
 
-	return map[string]string{"name": name}, nil
+		u.RawQuery = query.Encode()
+		req, _ := http.NewRequest(http.MethodGet, u.String(), nil)
+		rr, err := utils.MakeHTTPRequest(req)
+		if err != nil {
+			return nil, fmt.Errorf("unable to look up contact data:%s\n%s", err, rr.Response)
+		}
```
norkans7, line 676: I think it would be better to avoid repeating these lines below

so first if block to set the fields and one last if block to make the name from the response then always have the same last line as `return map[string]string{"name": name}, nil`

norkans7 (approved):
Looks good, thanks

PR #425, opened 2022-02-25
Update to support D3 contact check for whatsapp contact not in contact…

norkans7:
… store

### DIFF
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -1046,7 +1046,7 @@ func hasTiersError(payload mtErrorPayload) bool {
 
 func hasWhatsAppContactError(payload mtErrorPayload) bool {
 	for _, err := range payload.Errors {
-		if err.Code == 1006 && err.Title == "Resource not found" && err.Details == "unknown contact" {
+		if err.Code == 1006 && err.Title == "Resource not found" && (err.Details == "unknown contact" || err.Details == "Could not retrieve phone number from contact store") {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


handlers/whatsapp/whatsapp.go
```diff
@@ -1046,7 +1046,7 @@ func hasTiersError(payload mtErrorPayload) bool {
 
 func hasWhatsAppContactError(payload mtErrorPayload) bool {
 	for _, err := range payload.Errors {
-		if err.Code == 1006 && err.Title == "Resource not found" && err.Details == "unknown contact" {
+		if err.Code == 1006 && err.Title == "Resource not found" && (err.Details == "unknown contact" || err.Details == "Could not retrieve phone number from contact store") {
```
norkans7, line 1049: We checking the exact string since the same error code and error title are returned for missing file such as attachment so the details are the only place to make sure we need to check/revalidate the contact from WA API 

WA uses `unknown contact` as details and D3 uses `Could not retrieve phone number from contact store`

rowanseymour: Seems we shouldn't be checking error message strings no?.. error messages change, error codes should be stable

rowanseymour: wouldn't the most robust way to check this be `IF channel = D3 and error code = 1006` ?

norkans7: I am pretty sure D3 will also return 1006 for the file not found case as the wrap the WA API

norkans7: Also the details change could be something we the recent WA docker we have not yet tried yet

rowanseymour (approved):

PR #432, opened 2022-04-11
Add library with greater support for detection of mime types in Whatsapp

Robi9:
The added [library](https://github.com/gabriel-vasile/mimetype) has more type support than the currently used library, including `.xlsx` files that were previously being shipped as `.zip` due to lack of support from the current library.

### DIFF
--- a/handlers/whatsapp/whatsapp.go
+++ b/handlers/whatsapp/whatsapp.go
@@ -12,11 +12,11 @@ import (
 	"time"
 
 	"github.com/buger/jsonparser"
-	"github.com/gabriel-vasile/mimetype"
 	"github.com/gomodule/redigo/redis"
 	"github.com/nyaruka/courier"
 	"github.com/nyaruka/courier/handlers"
 	"github.com/nyaruka/courier/utils"
+	"github.com/nyaruka/gocommon/httpx"
 	"github.com/nyaruka/gocommon/urns"
 	"github.com/nyaruka/redisx"
 	"github.com/patrickmn/go-cache"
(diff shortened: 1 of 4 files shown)


handlers/whatsapp/whatsapp.go
```diff
@@ -863,7 +864,14 @@ func (h *handler) fetchMediaID(msg courier.Msg, mimeType, mediaURL string) (stri
 		return "", logs, errors.Wrapf(err, "error building request to media endpoint")
 	}
 	setWhatsAppAuthHeader(&req.Header, msg.Channel())
-	req.Header.Add("Content-Type", http.DetectContentType(rr.Body))
+	mtype := http.DetectContentType(rr.Body)
```
rowanseymour, line 867: is there any reason to call this first rather than just `mimetype.Detect(rr.Body)` ?

Robi9: This is the current library, the intention would be not to stop using it, but to add more support with the new library.

rowanseymour: Yes but what is the point of calling `http.DetectContentType(rr.Body)` first? Is there a case where it returns something correct that `mimetype.Detect(rr.Body)` doesn't ?

Robi9: Not that I have identified, I left it before because it is the current library and works well for other cases.

rowanseymour: I would assume that this new library detects a superset of things the built-in `http` function does, in which case let's just use it.

rowanseymour: Ok I added https://github.com/nyaruka/gocommon/pull/60 with some tests to see what the differences are - so let's update to gocommon v1.18.0 and anywhere we're calling `http.DetectContentType` we can call `httpx.DetectContentType` instead

Robi9: Ok, so I update this pull request, or would you prefer I open another one with the `gocommon v1.18.0` update and the function replacement for `httpx.DetectContentType`?

rowanseymour: Whatever is easiest. I already made the change in mailroom in https://github.com/nyaruka/mailroom/commit/dffc4980bcf948f5c36ee12d1869a0cbdc6d9164

PR #434, opened 2022-04-20
Quick fix to allow JSON content to not be omitted in logs

norkans7:
This is a quick fix allow JSON content in channel logs currently being omitted.

The proper refactor is to address https://github.com/nyaruka/courier/issues/435

### DIFF
--- a/channel_log.go
+++ b/channel_log.go
@@ -45,7 +45,7 @@ func sanitizeBody(body string) string {
 	ct := httpx.DetectContentType([]byte(parts[1]))
 
 	// if this isn't text, replace with placeholder
-	if !strings.HasPrefix(ct, "text") {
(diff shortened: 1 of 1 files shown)


channel_log.go
```diff
@@ -45,7 +45,7 @@ func sanitizeBody(body string) string {
 	ct := httpx.DetectContentType([]byte(parts[1]))
 
 	// if this isn't text, replace with placeholder
-	if !strings.HasPrefix(ct, "text") {
+	if !strings.HasPrefix(ct, "text") && ct != "application/json" {
```
rowanseymour, line 48: maybe better to use HasPrefix in case it gives us something like `application/json; charset=utf-8`

rowanseymour (approved):

Issue #433, opened 2022-04-14
Some contacts are being "duplicated" in orgs with whatsapp channel


rasoro:
scenery:
if you add a contact with whatsapp identifier 5582**9**99448232 and send a template message to that contact, when he replies a new contact with whatsapp identifier 558299448232 is created, because his `wa_id` does not have this additional 9.

context:
here in brazil in 2016 all mobile phone numbers needed to have this additional 9 to make calls, and many whatsapp contacts here created before that date have their wa_id without this 9.

suggestion:
I think when sending a message to a contact, we could check after request response if URN.path is different from the wa_id, if so update the contact's URN appropriately.

observation of a message being sent to this contact:
![Captura de tela de 2022-04-14 11-28-56](https://user-images.githubusercontent.com/75167930/163411954-f800de4a-d0f5-4327-b129-efd1ddf3734f.png)

matmsa27:
Hi @rowanseymour , what are your thoughts about this?

rowanseymour:
This is a tricky one but it does seem like we should be storing the version _without the extra 9_ if that is the contact's actual WhatsApp ID, even tho that then makes for [an invalid phone number](https://libphonenumber.appspot.com/phonenumberparser?number=%2B558299448232&country=BR).

And since contacts will always give their "WhatsApp number" with the extra 9 we need to fix that after the fact. Does sounds like we need to fix it when sending.

I'm not the most familiar with the WhatsApp API.. can they also be fixed in the `refresh_whatsapp_contacts` task?

@norkans7 @ericnewcomer

norkans7:
When we send a message to a contact and that fails we check the contact

https://github.com/nyaruka/courier/blob/5834bc2fb9789326d201b25ec373ab9e599bfb02/handlers/whatsapp/whatsapp.go#L953-L966

I am curious to know which number used when the contact replies to the channel, I would expect WA to send MO message using the wa_ID 558299448232 which is the one the accept us sending with.

Otherwise I think if whatsapp is sending in with 5582999448232 but only accept us sending with 558299448232 only, we always create a second contact for the incoming, not sure how to address this as we need to sort of have both at the same time.

Also we do not rely on the phone numbers for WhatsApp URNs

rasoro:
When we send a message to contact the extra 9 on the number, it doesn't fail.

But when the contact sends a message to the channel, the event that arrives at the courier is handled based on the URN assembled from the wa_id of the contact who sent it. As it does not find the contact, a new one is created.

[...]

PR #437, opened 2022-04-22
tweak in whatsapp handler on sendWhatsAppMsg returned wppID

rasoro:
This is a fix to update the contact URN if the wa_id returned in the send message request is different from the current URN path, avoiding creating a new contact.

I believe this is necessary in the following scenario.

Imagine a flow asking for the contact's whatsapp number through another channel type, as soon as the contact enters the number the flow sends a whatsapp template message to the contact successfully and the contact responds instantly.
If the whatsapp number is a valid phone number and whatsapp api accept the request but the wa_id is not the same as the number(as I mentioned here https://github.com/nyaruka/courier/issues/433), this will generate a new contact when the contact replies to the template message, thus "breaking" the flow for the initial contact.


handlers/whatsapp/whatsapp.go
```diff
@@ -1025,6 +1026,10 @@ func sendWhatsAppMsg(rc redis.Conn, msg courier.Msg, sendPath *url.URL, payload
 		return wppID, externalID, []*courier.ChannelLog{log, checkLog, retryLog}, err
 	}
 	externalID, err := getSendWhatsAppMsgId(rr)
+	wppID, err := jsonparser.GetString(rr.Body, "contacts", "[0]", "wa_id")
```
rowanseymour, line 1029: shouldn't do this if `err != nil` from line above

rowanseymour (changes requested):

handlers/whatsapp/whatsapp.go
rasoro: very important, thanks.

rasoro:
Any suggestions on how to increase coverage?

handlers/whatsapp/whatsapp_test.go
```diff
@@ -373,19 +373,19 @@ var defaultSendTestCases = []ChannelSendTestCase{
...
 		Status: "W", ExternalID: "157b5e14568e8",
-		ResponseBody: `{ "messages": [{"id": "157b5e14568e8"}] }`, ResponseStatus: 201,
+		ResponseBody: `{ "contacts":[{"input":"250788123123","wa_id":"250788123123"}], "messages": [{"id": "157b5e14568e8"}] }`, ResponseStatus: 201,
 		RequestBody: `{"to":"250788123123","type":"text","text":{"body":"Simple Message"}}`,
 		SendPrep:    setSendURL},
 	{Label: "Unicode Send",
 		Text: "☺", URN: "whatsapp:250788123123",
 		Status: "W", ExternalID: "157b5e14568e8",
-		ResponseBody: `{ "messages": [{"id": "157b5e14568e8"}] }`, ResponseStatus: 201,
+		ResponseBody: `{ "contacts":[{"input":"250788123123","wa_id":"250788123123"}], "messages": [{"id": "157b5e14568e8"}] }`, ResponseStatus: 201,
```
norkans7, line 388: Currently in our logs I see the responses we get are 

```
{"messages":[{"id":"gBGHI0gDaJdFfwIJIN2QOPc_Il5D"}],"meta":{"api_status":"stable","version":"2.37.2"}}
```

So we should not change these existing tests to have the contacts 

```
"contacts":[{"input":"250788123123","wa_id":"250788123123"}],
```

rasoro: I agree, thank you.

norkans7 (approved):

rowanseymour (approved):

PR #448, opened 2022-07-18
handling whatsapp cloud extra 9 contact number

rasoro:
This PR is related to this #437 which was also discussed here #433


norkans7:
I wondering if we really need to use this as now we can just send by phone number

https://developers.facebook.com/docs/whatsapp/cloud-api/reference/messages#message-object

and soon no WA ID will be used at all even for on-prem https://developers.facebook.com/docs/whatsapp/on-premises/reference/contacts

Was the issue reproduced with WhatsApp Cloud before adding this PR?

PR #449, opened 2022-07-19
Add link preview support in WAC

Robi9:
Support to include a URL preview in messages with links sent in Whatsapp. Add `"preview_url": true` to the Message Text Object as indicated in the [documentation](https://developers.facebook.com/docs/whatsapp/cloud-api/reference/messages#:~:text=of%20type%20video.-,Text%20Object,-Name).

### DIFF
--- a/handlers/facebookapp/facebookapp.go
+++ b/handlers/facebookapp/facebookapp.go
@@ -983,7 +983,7 @@ type wacComponent struct {
 
 type wacText struct {
 	Body       string `json:"body"`
-	PreviewURL bool   `json:"preview_url,omitempty"`
+	PreviewURL bool   `json:"preview_url"`
 }
 
 type wacLanguage struct {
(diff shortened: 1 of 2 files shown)


handlers/facebookapp/facebookapp.go
```diff
@@ -982,7 +982,8 @@ type wacComponent struct {
 }
 
 type wacText struct {
-	Body string `json:"body"`
+	Body       string `json:"body"`
+	PreviewURL bool   `json:"preview_url,omitempty"`
```
norkans7, line 986: Why not always have that explicitly set instead of omitting that

Robi9: If it is not sent it will be false by default so I thought of omitting it when I don't need to send it.

Robi9: Do you think it's better to make it explicit?

handlers/facebookapp/facebookapp.go
```diff
@@ -1086,8 +1086,14 @@ func (h *handler) sendCloudAPIWhatsappMsg(ctx context.Context, msg courier.Msg)
 			} else {
 				if i < (len(msgParts) + len(msg.Attachments()) - 1) {
 					// this is still a msg part
+					text := &wacText{}
```
norkans7, line 1089: text := &wacText{PreviewURL: false}

norkans7 (approved):

PR #451, opened 2022-07-20
Fix receiving attachments in WAC

Robi9:
Receiving attachments is not working properly. First, the token used to fetch the media url, channel auth_token, does not exist as WAC channels do not have this token in the channel settings. The token that must be used is the `whatsapp_admin_system_user_token` and in addition, for the media to be downloaded, it is necessary to use the same token.

### DIFF
--- a/backends/rapidpro/msg.go
+++ b/backends/rapidpro/msg.go
@@ -247,6 +247,11 @@ func downloadMediaToS3(ctx context.Context, b *backend, channel courier.Channel,
 		}
 	}
 
+	// to allow download media from wac when receive media message
+	if channel.ChannelType() == "WAC" {
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", b.config.WhatsappAdminSystemUserToken))
+	}
+
 	resp, err := utils.GetHTTPClient().Do(req)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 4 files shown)


handlers/facebookapp/facebookapp.go
```diff
@@ -424,21 +425,21 @@ func (h *handler) processCloudWhatsAppPayload(ctx context.Context, channel couri
 					text = msg.Text.Body
 				} else if msg.Type == "audio" && msg.Audio != nil {
 					text = msg.Audio.Caption
-					mediaURL, err = resolveMediaURL(channel, msg.Audio.ID)
+					mediaURL, err = resolveMediaURL(channel, msg.Audio.ID, token)
```
norkans7, line 428: I think channel is not needed in this case, so remove that parameter

norkans7 (approved):

backends/rapidpro/msg.go
```diff
@@ -247,6 +247,11 @@ func downloadMediaToS3(ctx context.Context, b *backend, channel courier.Channel,
 		}
 	}
 
+	// to allow download media from wac when receive media message
+	if channel.ChannelType() == "WAC" {
```
rowanseymour, line 251: backend shouldn't know about specific channel types - can't this be set in `BuildDownloadMediaRequest` ?

rowanseymour (changes requested):

norkans7 (approved):

rowanseymour (approved):

PR #460, opened 2022-08-02
Switch to `null.Map`

rowanseymour:


backends/rapidpro/channel.go
```diff
@@ -353,12 +349,7 @@ func (c *DBChannel) HasRole(role courier.ChannelRole) bool {
 
 // ConfigForKey returns the config value for the passed in key, or defaultValue if it isn't found
 func (c *DBChannel) ConfigForKey(key string, defaultValue interface{}) interface{} {
-	// no value, return our default value
-	if !c.Config_.Valid {
```
rowanseymour, line 357: we should convert fields like Channel.config and Org.config to be actual JSON so there's no danger of them becoming invalid

backends/rapidpro/channel.go
```diff
@@ -50,32 +50,30 @@ func getChannel(ctx context.Context, db *sqlx.DB, channelType courier.ChannelTyp
 	return channel, nil
 }
 
-const lookupChannelFromUUIDSQL = `
-SELECT 
-	org_id, 
-	ch.id as id,
```
rowanseymour, line 56: removing redundant aliasing

norkans7 (approved):

PR #462, opened 2022-08-03
Add handler support for resolving media

rowanseymour:
See https://github.com/rapidpro/rapidpro/issues/1679. So handlers can call...

```go
var mediaSupport = map[handlers.MediaType]handlers.MediaTypeSupport{
	handlers.MediaTypeImage:       {Types: []string{"image/jpeg"}, MaxBytes: 10 * 1024 * 1024},
	handlers.MediaTypeApplication: {Types: []string{"application/pdf"}, MaxBytes: 50 * 1024 * 1024},
}

handlers.ResolveAttachments(ctx, h.Backend(), msg.Attachments(), mediaSupport, false)
```

And they get back an array of `Attachment` objects...

```go
type Attachment struct {
	Type        MediaType
	ContentType string
	URL         string
	Media       courier.Media
	Thumbnail   courier.Media
}
```

Which will always have `Type` (e.g. "image" etc), `ContentType` (e.g. "image/jpeg") and URL set, and may additionally have `Media` and `Thumbnail` set.

If a channel knows it won't be able to support URL-only attachments, it can pass `allowURLOnly=false`


backends/rapidpro/backend.go
```diff
@@ -410,28 +410,18 @@ func (b *backend) WriteExternalIDSeen(msg courier.Msg) {
...
-	if len(parts) <= 1 || strings.HasPrefix(parts[1], "//") {
-		return nil, errors.Errorf("invalid attachment format: %s", a)
-	}
-
-	contentType, mediaUrl = parts[0], parts[1]
-	stub := &DBMedia{URL_: mediaUrl, ContentType_: contentType}
-
-	// if we can't parse the URL or the hostname isn't our media domain, return stub
+// ResolveMedia resolves the passed in attachment URL to a media object
+func (b *backend) ResolveMedia(ctx context.Context, mediaUrl string) (courier.Media, error) {
```
rowanseymour, line 414: moved attachment string (content-type:url) parsing out of backend and into handler util

backend.go
```diff
@@ -106,7 +105,6 @@ type Backend interface {
 
 // Media is a resolved media object that can be used as a message attachment
 type Media interface {
-	UUID() uuids.UUID
```
rowanseymour, line 109: backend implementation detail

handlers/telegram/telegram.go
```diff
@@ -187,9 +187,14 @@ func (h *handler) SendMsg(ctx context.Context, msg courier.Msg) (courier.MsgStat
 		return nil, fmt.Errorf("invalid auth token config")
 	}
 
+	attachments, err := handlers.ResolveAttachments(ctx, h.Backend(), msg.Attachments(), nil, true)
```
rowanseymour, line 190: I can't find a list of supported types for telegram.. so this will return all attachments regardless of type

handlers/rocketchat/rocketchat_test.go
```diff
@@ -1,10 +1,11 @@
 package rocketchat
 
 import (
-	"github.com/nyaruka/courier"
-	. "github.com/nyaruka/courier/handlers"
```
rowanseymour, line 5: this causes rocketchat.Attachment to conflict with handlers.Attachment

handlers/media.go
```diff
@@ -0,0 +1,150 @@
...
+	resolved := make([]*Attachment, 0, len(attachments))
+
+	for _, as := range attachments {
+		att, err := resolveAttachment(ctx, b, as, support, allowURLOnly)
+		if err != nil {
+			return nil, err
+		}
+		if att != nil {
+			resolved = append(resolved, att)
+		}
```
rowanseymour, line 45: Allowing attachments to be skipped over if they're not supported... debatable what we should do in this case? Could just error the message too..

rowanseymour: I guess skipping over attachments is at least consistent with how it works currently if a channel handler just does't support sending attachments.

norkans7: I thought when a media is over the limit of the max supported by a channel type we show a channel log with that error specified

norkans7: So telegram send an image well but WA fail because that image is bigger than the max for WA

norkans7: Here I think we are silently skipping the attachment

norkans7: @rowanseymour have you seen this?

rowanseymour: Yeah.. I'm not keen on adding more reasons to create channel logs, until we can rework channel logs in general.

I appreciate this might be confusing but it's no more confusing than the current behavior which is silently skipping if the channel type doesn't support attachments.

norkans7: I was thinking now we try to send the the channel API the file they do not support and then that is rejected and we have a log now that we see the response form the API has rejected the file type.

Now here I think we are not at all making the request and fail and we silently skip the attachment.

My understanding is that we can create the log without making the API request, but yes we can add that later when the logs are reworked.

handlers/media.go
```diff
@@ -0,0 +1,170 @@
...
+	if len(mediaSupport.Types) > 0 {
+		candidates = filterMediaByContentTypes(candidates, mediaSupport.Types)
+	}
+
+	// narrow down the candidates to the ones that don't exceed our max bytes
+	if mediaSupport.MaxBytes > 0 {
+		candidates = filterMediaBySize(candidates, mediaSupport.MaxBytes)
+	}
+
+	// if we have no candidates, we can't use this media
```
norkans7, line 100: I think we need to be backward compatible for the attachments uploaded before we support media as DB objects too

rowanseymour: I think that has to depend on the channel type - for some channel types it's fine if we have a URL we can't match to a media DB object. For LINE it won't work but I'm fine telling our LINE users to re-upload attachments. I don't think we can automatically backfill every attachment ever uploaded.

rowanseymour: In the LINE channel I think we resolve attachments with `allowURLOnly = false` and we send whatever attachments we get back which will have an associated `Media` object.

norkans7 (approved):

PR #464, opened 2022-08-05
Fix URLs of attachments in WAC handler

Robi9:
Files where the name has spaces or some special character end up not being sent because their link is not being encoded to correct this.


rowanseymour:
The problem here isn't specific to WAC channels so I'd like to address this instead in https://github.com/nyaruka/courier/pull/462 and then migrate all channels to use `handlers.ResolveAttachments`

PR #474, opened 2022-08-12
Support media attachments for LINE

norkans7:


handlers/line/line.go
```diff
@@ -32,6 +32,14 @@ var (
 	signatureHeader = "X-Line-Signature"
 )
 
+// see https://core.telegram.org/bots/api#sending-files
+var mediaSupport = map[handlers.MediaType]handlers.MediaTypeSupport{
+	handlers.MediaTypeImage:       {Types: []string{"image/jpe", "image/jpg", "image/jpeg", "image/png"}, MaxBytes: 10 * 1024 * 1024},
+	handlers.MediaTypeAudio:       {Types: []string{"audio/mp4", "audio/m4a"}, MaxBytes: 200 * 1024 * 1024},
```
rowanseymour, line 38: on the rapidpro side we always correct `audio/m4a` to `audio/mp4` so don't think this is needed

handlers/line/line.go
```diff
@@ -32,6 +32,14 @@ var (
 	signatureHeader = "X-Line-Signature"
 )
 
+// see https://core.telegram.org/bots/api#sending-files
+var mediaSupport = map[handlers.MediaType]handlers.MediaTypeSupport{
+	handlers.MediaTypeImage:       {Types: []string{"image/jpe", "image/jpg", "image/jpeg", "image/png"}, MaxBytes: 10 * 1024 * 1024},
```
rowanseymour, line 37: I don't think I've ever seen a browser upload something as `image/jpe`.. I'm inclined to say if there's a chance that can happen we should have rapidpro correct `image/jpe` to `image/jpeg` rather than have every channel handler have to worry about `image/jpe` vs `image/jpg` vs `image/jpeg`

handlers/line/line.go
```diff
@@ -300,19 +320,29 @@ func (h *handler) Send(ctx context.Context, msg courier.Msg, logger *courier.Cha
 			}
 		}
 	}
+
+	attachments, err := handlers.ResolveAttachments(ctx, h.Backend(), msg.Attachments(), mediaSupport, true)
```
rowanseymour, line 324: `allowURLOnly` should be false otherwise this can return attachments that don't have associated media objects

rowanseymour (changes requested):

handlers/line/line.go
```diff
@@ -32,6 +32,14 @@ var (
 	signatureHeader = "X-Line-Signature"
 )
 
+// see https://core.telegram.org/bots/api#sending-files
```
rowanseymour, line 35: update comment to be LINE api docs

rowanseymour (changes requested):

handlers/line/line.go
norkans7: updated

handlers/line/line.go
norkans7: Added a test PDF media and enforced that

handlers/line/line.go
norkans7: Updated

handlers/line/line.go
norkans7: done

rowanseymour (approved):

PR #478, opened 2022-08-16
Adjust to support sending attachments with quick replies later

norkans7:


handlers/line/line.go
```diff
@@ -347,6 +323,31 @@ func (h *handler) Send(ctx context.Context, msg courier.Msg, logger *courier.Cha
 			jsonMsgs = append(jsonMsgs, string(jsonMsg))
 		}
 	}
+
```
norkans7, line 326: Move to send text messages after all attachments

rowanseymour (approved):

Issue #456, opened 2022-07-29
Support for sending attachments with captions


Robi9:
Hello everyone, I would like us to discuss the way in which the sending of attachments with text in the `WAC` type channel is handled.
We saw that the text of attachments can be sent as a caption, which could make it even more intuitive for the user. The way it is done today, sending the attachment first and then sending the text message, could be improved by adding the use of captions.
I know that the way currently used was thought about the maximum size of the message, but what if we did a check on the size of the text that we could send it as a caption if it is less than or equal to the maximum size of the text, or send it separately of the attachment, the way it is done today when it becomes larger than the maximum size.
I brought up these questions because we saw that this change will impact quick replies and lists replies that have text and attachment since the message sending payload assembly function in the handler was built based on these separate sends. Can we discuss this further?

matmsa27:
Hi, @rowanseymour and @norkans7 what are your thoughts about this improvement on whats app cloud handler?

norkans7:
Sounds good, please open a PR for that.
It is not clear to me how that affect the quick replies though or how the quick replies have attachments we'll see with those changes

PR #482, opened 2022-08-22
Add attachment caption support in WAC handler

Robi9:
As mentioned in https://github.com/nyaruka/courier/issues/456 we will do a validation to see if the text to be sent with the attachment fits within the size limit that WhatsApp assigns, if it does, it will be sent in the `caption` of the attachment, otherwise it will be sent first the attachment and then the text, as is currently done. Also added this feature when we have message with attachment, text and quick replies, they will be sent together in an `interactive` message.


norkans7:
@Robi9 these changes are hard to review as they are now mixed that break this into 2 PR one for caption and the interactive part in a second PR

Pay attention to the `omitempty` tag that seems to have been not taking effect because of the the type changes

Robi9:
@norkans7 PR for attachments with captions: https://github.com/nyaruka/courier/pull/494

PR #487, opened 2022-08-25
Handler tests should check for channel logs

rowanseymour:


handlers/test.go
```diff
@@ -49,8 +49,8 @@ type ChannelHandleTestCase struct {
 	ExpectedExternalID  *string
 	ExpectedMsgID       int64
 
-	ChannelEvent      *string
-	ChannelEventExtra map[string]interface{}
+	ExpectedChannelEvent      courier.ChannelEventType
```
rowanseymour, line 52: there's no valid channel_event type of "" so this doesn't have to be a pointer

PR #489, opened 2022-08-25
Redaction

rowanseymour:


handlers/facebookapp/facebookapp.go
```diff
@@ -29,8 +29,6 @@ var (
 
 	signatureHeader = "X-Hub-Signature"
 
-	configWACPhoneNumberID = "wac_phone_number_id"
```
rowanseymour, line 32: not used

handlers/blackmyna/blackmyna.go
```diff
@@ -144,3 +144,9 @@ func (h *handler) Send(ctx context.Context, msg courier.Msg, clog *courier.Chann
 
 	return status, nil
 }
+
+func (h *handler) RedactValues(ch courier.Channel) []string {
+	return []string{
+		handlers.BasicAuth(ch.StringConfigForKey(courier.ConfigUsername, ""), ch.StringConfigForKey(courier.ConfigPassword, "")),
+	}
```
norkans7, line 151: Don't we want to always include the default config keys? 
https://github.com/nyaruka/courier/pull/489/files#diff-81879b15fa52c9f649a3c866b7296b3c704339e8446b5f29ab98811e2dcb8c0bR11

So channel type handler will always append to the default keys

rowanseymour:
@norkans7 this PR enables redaction on incoming requests too, and I wanted to add testing of that, but I can't find an example of a channel passing back something considered sensitive, any ideas?

handlers/blackmyna/blackmyna.go
norkans7: that seems like a mistake `api_key` should not be there

norkans7:
Probably the logs from describeURN for an incoming message.

handlers/blackmyna/blackmyna.go
norkans7: I was think we always redact all those keys just in case the code is updates and one key is forgotten to be updated for those common one.

However this is fine if we keep the rapidpro redaction side as well at display time especially for the most sensitive info such WAC token

norkans7 (approved):

handlers/blackmyna/blackmyna.go
rowanseymour: I don't think we keep redaction on the RapidPro side - we don't want to maintain code that we assume is never used because there's nothing left to redact. However in the case of keys that belong to us (e.g. WAC) we can add an assert that they're not contained in any displayed log.

norkans7: I think we keep the rapidpro even for the anon related redaction

rowanseymour: The anon stuff we keep in RapidPro, but access key stuff is courier's responsibility.

If you want to be certain something is happening, you don't do it twice and hope for the best, but you find a way to assert that it's happening. If a user's access key ends up in one of their logs, that's not the end of the world - only they can see them. If one of our WAC keys ends up in there, that's a problem. So we'll assert that's not the case.

norkans7 (approved):

PR #498, opened 2022-09-06
Support channels receiving embedded attachments and use with thinq handler

rowanseymour:


backends/rapidpro/msg.go
```diff
@@ -262,7 +270,7 @@ func downloadMediaToS3(ctx context.Context, b *backend, channel courier.Channel,
 	}
 
 	// first try getting our mime type from the first 300 bytes of our body
-	fileType, err := filetype.Match(body[:300])
+	fileType, _ := filetype.Match(body[:300])
```
rowanseymour, line 273: err wasn't being checked

backends/rapidpro/msg.go
```diff
@@ -478,6 +491,12 @@ func writeExternalIDSeen(b *backend, msg courier.Msg) {
 // Our implementation of Msg interface
 //-----------------------------------------------------------------------------
 
+type embeddedAttachment struct {
```
norkans7, line 494: these won't be supported in case we write messages from the spool?
Can we make these serializable to JSON and support that?

rowanseymour: hmm good point...

rowanseymour: https://en.wikipedia.org/wiki/Data_URI_scheme

norkans7 (approved):

PR #501, opened 2022-09-07
Rework handler tests to assert more state by default

rowanseymour:


handlers/facebook/facebook_test.go
```diff
@@ -547,7 +547,6 @@ var testCases = []ChannelHandleTestCase{
 		Data:               dlr,
 		ExpectedRespStatus: 200,
 		ExpectedRespBody:   "Handled",
-		ExpectedDate:       time.Date(2016, 4, 7, 1, 11, 27, 970000000, time.UTC),
```
rowanseymour, line 550: statuses don't have dates

PR #502, opened 2022-09-08
Adjust to use the cache by address correctly

norkans7:


norkans7:
We have 2 ways to lookup channel one by uuid and the other by address for channel that use a common receive endpoint such as /c/fba/receive

I just noticed the caching was not correct for the lookup by address.

PR #507, opened 2022-09-14
Webhooks responses

norkans7:


server.go
```diff
@@ -274,7 +274,14 @@ func (s *server) channelHandleWrapper(handler ChannelHandler, handlerFunc Channe
 		// get the channel for this request - can be nil, e.g. FBA verification requests
 		channel, err := handler.GetChannel(ctx, r)
 		if err != nil {
-			logrus.WithError(err).WithField("request", string(recorder.Trace.RequestTrace))
+
+			// webhooks for FBA, IG and WAC should always return 200
+			if !handler.UseChannelRouteUUID() {
```
rowanseymour, line 279: I get that currently the 3 channels that want errors as 200s are the same 3 that don't use channel UUID routing... but feels like this could be more explicit... maybe handler defines a `ErrorResponseStatus() int` method and baseHandler returns 400?

rowanseymour: Could change `WriteRequestError` to take a status code, and then `WriteAndLogRequestError` should get the status code from the handler..

rowanseymour (changes requested):

server.go
```diff
@@ -274,7 +274,14 @@ func (s *server) channelHandleWrapper(handler ChannelHandler, handlerFunc Channe
 		// get the channel for this request - can be nil, e.g. FBA verification requests
 		channel, err := handler.GetChannel(ctx, r)
 		if err != nil {
-			logrus.WithError(err).WithField("request", string(recorder.Trace.RequestTrace))
+
+			// webhooks for FBA, IG and WAC should always return 200
+			if handler.ErrorResponseStatus() == 200 {
```
rowanseymour, line 279: if we want all places where we can return an error to use 200, then can we update `writeAndLogRequestError` to take the handler and call `ErrorResponseStatus()` ?

rowanseymour (changes requested):

server.go
norkans7: `writeAndLogRequestError` is in courier package and trying to reference handlers there causes cycle import issue

rowanseymour: Ok.. we can merge this to get FBA working.. but `handler.ErrorResponseStatus() == 200` is not the greatest bit of code ever written so we should take a pass at trying to refactor how handlers customize responses.

rowanseymour (approved):

PR #508, opened 2022-09-15
Update to use SHA256 signature for FBA payload

norkans7:


handlers/facebookapp/facebookapp.go
```diff
@@ -27,7 +27,9 @@ var (
 	sendURL  = "https://graph.facebook.com/v12.0/me/messages"
 	graphURL = "https://graph.facebook.com/v12.0/"
 
-	signatureHeader = "X-Hub-Signature"
+	signatureHeader = "X-Hub-Signature-256"
+
+	maxRequestBodyBytes int64 = 500000
```
norkans7, line 32: I increased that from 100000 as I noticed a request that is failing as is has more than 300000

rowanseymour: Let's do a MB ... (1024 *1024) We're not expecting Facebook to try to DDOS us. What was in the request that was 300KB?

rowanseymour (changes requested):

handlers/facebookapp/facebookapp.go
norkans7: so many "\u200a" and "*\u06d4\u06d4\u06e9\u06d4\u06d4*." but that length limit is causing us to not validate the signature properly and fail the request that is kept retrying

rowanseymour (approved):

PR #513, opened 2022-09-22
🧪 Better errors in channel logs

rowanseymour:
Thinking about these errors as being user facing... and making it easier to point users to resources on external sites, e.g. we get a 20002 from a Twilio call and we can turn that into a link to https://www.twilio.com/docs/api/errors/20002.

### DIFF
--- a/channel_log.go
+++ b/channel_log.go
@@ -1,6 +1,7 @@
 package courier
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/nyaruka/gocommon/dates"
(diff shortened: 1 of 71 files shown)


channel_log.go
```diff
@@ -34,6 +35,29 @@ func NewChannelError(message, code string) *ChannelError {
 	return &ChannelError{message: message, code: code}
 }
 
+func ErrorResponseStatusCode() *ChannelError {
+	return NewChannelError("Unexpected response status code.", "core:response_status_code")
```
rowanseymour, line 39: don't need to repeat the status code in the error message because they can see it in the trace above

handlers/dart/dart.go
```diff
@@ -184,16 +185,16 @@ func (h *handler) Send(ctx context.Context, msg courier.Msg, clog *courier.Chann
...
+				errorMessage = "Authentication error."
+			} else if responseCode == "101" {
+				errorMessage = "Account expired or invalid parameters."
 			}
-			if responseText == "101" {
-				errorMessage = "Error 101: Account expired or invalid parameters"
-			}
-			clog.Error(fmt.Errorf(errorMessage))
+
+			clog.Error(courier.ErrorServiceSpecific("dart", responseCode, errorMessage))
```
rowanseymour, line 197: Here we know what some of the error codes mean so we can generate some error messages.. but we want to avoid having to interpret all the possible error codes we can get back from a provider like Twilio, so we can just pass an empty message to use a generic one, and then point users to the external site.

norkans7 (approved):

PR #519, opened 2022-10-04
Add endpoint to download and store attachments by their URL

rowanseymour:
Still pondering on how mailroom accesses this

### DIFF
--- a/test/backend.go
+++ b/test/backend.go
@@ -328,6 +328,11 @@ func (mb *MockBackend) WriteExternalIDSeen(msg courier.Msg) {
 	mb.seenExternalIDs = append(mb.seenExternalIDs, msg.ExternalID())
 }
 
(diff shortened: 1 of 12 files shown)


test/backend.go
```diff
@@ -75,21 +88,6 @@ func NewMockBackend() *MockBackend {
...
-
-// LastContactName returns the contact name set on the last msg or channel event written
-func (mb *MockBackend) LastContactName() string {
-	return mb.lastContactName
-}
-
-// MockMedia adds the given media to the mocked backend
-func (mb *MockBackend) MockMedia(media courier.Media) {
-	mb.media[media.URL()] = media
-}
```
rowanseymour, line 91: moving these methods which aren't actually part of the Backend interface down to the bottom in their own section

norkans7 (approved):

PR #523, opened 2022-10-06
Server and responses tweaks

rowanseymour:


server.go
```diff
@@ -62,15 +62,15 @@ func NewServerWithLogger(config *Config, backend Backend, logger *logrus.Logger)
 	router.Use(middleware.Recoverer)
 	router.Use(middleware.Timeout(30 * time.Second))
 
-	chanRouter := chi.NewRouter()
-	router.Mount("/c/", chanRouter)
+	publicRouter := chi.NewRouter()
```
rowanseymour, line 65: because we're going to add at least one point that isn't a channel to this

PR #524, opened 2022-10-07
Rework attachment fetching to keep URL and content type separate

rowanseymour:
It's convenient to store attachments as a single string but that only needs to be a last minute thing when writing to the db

### DIFF
--- a/backends/rapidpro/task.go
+++ b/backends/rapidpro/task.go
@@ -16,9 +16,9 @@ func queueMsgHandling(rc redis.Conn, c *DBContact, m *DBMsg) error {
 	body := map[string]any{
 		"contact_id":      c.ID_,
 		"org_id":          channel.OrgID_,
-		"channel_id":      channel.ID_, // deprecated
-		"channel_uuid":    channel.UUID_,
-		"channel_type":    channel.ChannelType_,
+		"channel_id":      channel.ID_,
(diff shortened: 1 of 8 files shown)


backends/rapidpro/task.go
```diff
@@ -16,9 +16,9 @@ func queueMsgHandling(rc redis.Conn, c *DBContact, m *DBMsg) error {
 	body := map[string]any{
 		"contact_id":      c.ID_,
 		"org_id":          channel.OrgID_,
-		"channel_id":      channel.ID_, // deprecated
-		"channel_uuid":    channel.UUID_,
-		"channel_type":    channel.ChannelType_,
+		"channel_id":      channel.ID_,
+		"channel_uuid":    channel.UUID_,        // deprecated
+		"channel_type":    channel.ChannelType_, // deprecated
```
rowanseymour, line 21: I was thinking mailroom would need these but the first thing mailroom does is grab that channel by its id and that then has the UUID and type

norkans7: Just so we remember reverting https://github.com/nyaruka/rapidpro/pull/4033/files as well

rowanseymour: Ayyyy I totally forgot well remembered!

rowanseymour: https://github.com/nyaruka/rapidpro/pull/4040

PR #529, opened 2022-10-17
Remove no longer used `channel_uuid` and `channel_type` fields from msg events

rowanseymour:
In the end mailroom just uses `channel_id`


backends/rapidpro/backend_test.go
```diff
@@ -1152,7 +1150,6 @@ func (ts *BackendTestSuite) TestWriteMsg() {
 		"text":            msg.Text(),
 		"attachments":     nil,
 		"new_contact":     contact.IsNew_,
-		"created_on":      msg.CreatedOn_.Format(time.RFC3339Nano),
```
rowanseymour, line 1155: also not used https://github.com/nyaruka/mailroom/pull/679

PR #532, opened 2022-10-19
Add support for JustCall channel type

norkans7:


handlers/justcall/justcall.go
```diff
@@ -0,0 +1,239 @@
...
+
+	resp, respBody, err := handlers.RequestHTTP(req, clog)
+	if err != nil || resp.StatusCode/100 != 2 {
+		return status, nil
+	}
+
+	respStatus, err := jsonparser.GetString(respBody, "status")
+	if err != nil {
+		clog.Error(courier.ErrorResponseValueMissing("status"))
+		clog.End()
```
rowanseymour, line 217: handlers don't need to call `clog.End()` - the sender does this

handlers/justcall/justcall.go
```diff
@@ -0,0 +1,239 @@
...
+	} else {
+		text = handlers.GetTextAndAttachments(msg)
+	}
+
+	payload := mtPayload{From: msg.Channel().Address(), To: msg.URN().Path(), Body: text}
+	if len(mediaURLs) > 0 {
+		payload.MediaURL = strings.Join(mediaURLs, ",")
+	}
+
+	jsonPayload, err := json.Marshal(payload)
```
rowanseymour, line 196: I think it's ok to use `jsonx.MustMarshal` in cases like this since we control everything in that payload and we know it can be marshalled

handlers/justcall/justcall.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	if respStatus != "success" {
+		return status, nil
+
+	}
+
+	externalID, err := jsonparser.GetInt(respBody, "id")
+	if err != nil {
+		clog.Error(courier.ErrorResponseValueMissing("id"))
+		clog.End()
```
rowanseymour, line 228: same here

rowanseymour (changes requested):

rowanseymour (approved):

PR #535, opened 2022-10-26
🔐 Fetch access token for JioChat and WeChat channels as needed

rowanseymour:
Instead of relying on scheduled task in RP

Also fixes channel log redaction for these channel types.

### DIFF
--- a/handlers/jiochat/jiochat.go
+++ b/handlers/jiochat/jiochat.go
@@ -232,6 +232,12 @@ func (h *handler) DescribeURN(ctx context.Context, channel courier.Channel, urn 
 	return map[string]string{"name": nickname}, nil
 }
 
+func (h *handler) RedactValues(ch courier.Channel) []string {
+	return []string{
+		ch.StringConfigForKey(configAppSecret, ""),
+	}
+}
+
 // BuildAttachmentRequest download media for message attachment
 func (h *handler) BuildAttachmentRequest(ctx context.Context, b courier.Backend, channel courier.Channel, attachmentURL string, clog *courier.ChannelLog) (*http.Request, error) {
 	parsedURL, err := url.Parse(attachmentURL)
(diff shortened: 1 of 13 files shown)


handlers/jiochat/jiochat.go
```diff
@@ -41,10 +40,15 @@ func init() {
 
 type handler struct {
 	handlers.BaseHandler
+
+	fetchTokenMutex sync.Mutex
```
rowanseymour, line 44: synchronizes all jiochat token lookups but we don't do a lot of traffic over jiochat and it's almost always a single redis GET so not too worried

handlers/jiochat/jiochat.go
```diff
@@ -161,71 +161,6 @@ func buildMediaURL(mediaID string) string {
...
-		clog.Error(courier.ErrorResponseValueMissing("access_token"))
-		clog.End()
-		return h.Backend().WriteChannelLog(ctx, clog)
-	}
-
-	rc := h.Backend().RedisPool().Get()
-	defer rc.Close()
-
-	cacheKey := fmt.Sprintf("jiochat_channel_access_token:%s", channel.UUID().String())
-	_, err = rc.Do("SET", cacheKey, accessToken, 7200)
```
rowanseymour, line 205: this has always been broken because this returns a redis syntax error.. should be `rc.Do("SET", cacheKey, accessToken, "EX", 7200)`

norkans7 (commented):
I think is is best to move similar wechat refresh task here too

handlers/jiochat/jiochat_test.go
```diff
@@ -18,13 +16,14 @@ import (
...
 	. "github.com/nyaruka/courier/handlers"
 	"github.com/nyaruka/courier/test"
+	"github.com/nyaruka/gocommon/httpx"
 	"github.com/nyaruka/gocommon/urns"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
 var testChannels = []courier.Channel{
-	test.NewMockChannel("8eb23e93-5ecb-45ba-b726-3b064e0c56ab", "JC", "2020", "US", map[string]interface{}{configAppSecret: "secret", configAppID: "app-id"}),
```
rowanseymour, line 27: tweaked this because `secret` occurs as a field in the payload for wechat so it's confusing if it's also the value that should be redacted...

norkans7 (approved):

PR #546, opened 2022-11-21
Add logs for WhatsApp Cloud specific errors

norkans7:


handlers/facebookapp/facebookapp.go
```diff
@@ -439,6 +447,11 @@ func (h *handler) processCloudWhatsAppPayload(ctx context.Context, channel couri
 					return nil, nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r, err)
 				}
 
+				for _, msg_error := range msg.Errors {
```
rowanseymour, line 450: hmm writing some Python here hehe?

rowanseymour (changes requested):

handlers/facebookapp/facebookapp.go
norkans7: updated the variables name

PR #550, opened 2022-11-24
Add support for external telegram errors

norkans7:


handlers/telegram/telegram_test.go
```diff
@@ -613,7 +677,23 @@ var testCases = []ChannelHandleTestCase{
...
+		ExpectedBodyContains: "unable to resolve file",
+		ExpectedErrors:       []*courier.ChannelError{courier.ErrorResponseUnparseable("JSON")},
+	},
+	{
+		Label:                "Receive error File response",
+		URL:                  "/c/tg/8eb23e93-5ecb-45ba-b726-3b064e0c568c/receive/",
+		Data:                 errorFile,
+		ExpectedRespStatus:   200,
+		ExpectedBodyContains: "unable to resolve file",
+		ExpectedErrors:       []*courier.ChannelError{courier.ErrorResponseUnparseable("JSON")},
```
rowanseymour, line 696: isn't this test case supposed to test catching an error code?

norkans7: we are parsing the response into struct now so if 'ok' is not present the struct will have that as false so we catch both the same way

norkans7: the JSON error we catch now are new tests I added

rowanseymour (approved):

PR #557, opened 2023-01-23
Remove passing the parameters as null for WA template components

norkans7:


norkans7:
We should wait a little to merge this

rowanseymour (approved):

handlers/whatsapp/whatsapp.go
```diff
@@ -436,7 +436,7 @@ type Param struct {
 
 type Component struct {
 	Type       string  `json:"type"`
-	Parameters []Param `json:"parameters"`
```
rowanseymour, line 439: or we can make sure this is initialized to an empty array so sends `[]`

norkans7: Yeah, I had thought of that but went with interpreting optional as something that is present or not

PR #559, opened 2023-01-26
Support reading `locale` instead of `language`+`country` on templating

rowanseymour:


handlers/facebookapp/facebookapp.go
```diff
@@ -1568,23 +1557,34 @@ func (h *handler) BuildAttachmentRequest(ctx context.Context, b courier.Backend,
...
+
+	// if we have a country, strip that off and look again for a match
+	l, c := lc.ToParts()
+	if c != "" {
+		if lang := supportedLanguages[courier.Locale(l)]; lang != "" {
+			return lang
+		}
+	}
+	return "en" // fallback to English
+}
```
rowanseymour, line 1583: there 2 changes to the logic here...

 1. If we're given a code like `swa-KE`, we try that, but also try `swa`
 2. If we don't find a match, fallback on `en`... this seems to me preferable to failing

@ericnewcomer @norkans7 thoughts?

handlers/facebookapp/facebookapp_test.go
```diff
@@ -1526,3 +1534,13 @@ func TestBuildMediaRequest(t *testing.T) {
...
 }
+
+func TestGetSupportedLanguage(t *testing.T) {
+	assert.Equal(t, "en", getSupportedLanguage(courier.NilLocale))
+	assert.Equal(t, "en", getSupportedLanguage(courier.Locale("eng")))
+	assert.Equal(t, "en_US", getSupportedLanguage(courier.Locale("eng-US")))
+	assert.Equal(t, "pt_PT", getSupportedLanguage(courier.Locale("por")))
+	assert.Equal(t, "pt_PT", getSupportedLanguage(courier.Locale("por-PT")))
+	assert.Equal(t, "pt_BR", getSupportedLanguage(courier.Locale("por-BR")))
+	assert.Equal(t, "fil", getSupportedLanguage(courier.Locale("fil")))
```
norkans7, line 1545: we need an extra test for such `fr-CA` that should get `fr` too

norkans7: ignoring the country

handlers/facebookapp/facebookapp.go
norkans7: That makes sense

norkans7 (approved):
Looks good, just the test for coverage

PR #562, opened 2023-01-31
Update to v2 of nyaruka/null

rowanseymour:


backends/rapidpro/org.go
```diff
@@ -1,33 +1,4 @@
...
-// Value returns the db value, null is returned for 0
-func (i OrgID) Value() (driver.Value, error) {
-	return null.Int(i).Value()
-}
-
-// Scan scans from the db value. null values become 0
-func (i *OrgID) Scan(value interface{}) error {
-	return null.ScanInt(value, (*null.Int)(i))
-}
+type OrgID int
```
rowanseymour, line 4: null org IDs aren't a thing

norkans7 (approved):

PR #569, opened 2023-03-14
Reduce time allowed for attachment requests so that we return before server cancels us

rowanseymour:


utils/http.go
```diff
@@ -10,13 +10,13 @@ import (
...
 	once.Do(func() {
-		transport = http.DefaultTransport.(*http.Transport).Clone()
+		transport := http.DefaultTransport.(*http.Transport).Clone()
 		transport.MaxIdleConns = 64
 		transport.MaxIdleConnsPerHost = 8
 		transport.IdleConnTimeout = 15 * time.Second
 		client = &http.Client{
 			Transport: transport,
-			Timeout:   60 * time.Second,
+			Timeout:   30 * time.Second,
```
rowanseymour, line 19: this puts a 30 second limit on all outgoing requests from courier... which seems fine.. if you haven't returned in 30 seconds you're probably not going to return

server.go
```diff
@@ -126,7 +126,7 @@ func (s *server) Start() error {
 		Addr:         fmt.Sprintf("%s:%d", s.config.Address, s.config.Port),
 		Handler:      s.router,
 		ReadTimeout:  30 * time.Second,
-		WriteTimeout: 30 * time.Second,
+		WriteTimeout: 45 * time.Second,
```
rowanseymour, line 129: this has to be bigger than the 30 seconds alloted to request an attachment from a remote server

utils/http.go
norkans7: yes, 30 sec is still plenty of time

norkans7 (approved):

PR #572, opened 2023-03-16
MTN channel

norkans7:


handlers/mtn/mtn.go
```diff
@@ -0,0 +1,245 @@
...
+func (h *handler) receiveStatus(ctx context.Context, channel courier.Channel, w http.ResponseWriter, r *http.Request, clog *courier.ChannelLog) ([]courier.Event, error) {
+	payload := &statusPayload{}
+	err := handlers.DecodeAndValidateJSON(payload, r)
+	if err != nil {
+		return nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r, err)
+	}
+	msgStatus, found := statusMapping[payload.DeliveryStatus[0].Status]
+	if !found {
+		return nil, handlers.WriteAndLogRequestError(ctx, h, channel, w, r,
+			fmt.Errorf("unknown status '%s', must be one of 'DeliveredToTerminal','DeliveryUncertain','DeliveryImpossible','DeliveredToNetwork', 'MessageWaiting', or 'DeliveryNotificationNotSupported'", payload.DeliveryStatus[0].Status))
```
rowanseymour, line 81: I don't think you need to list all statuses here.. this error message is only for us to debug and we know what the accepted statuses are

handlers/mtn/mtn.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (h *handler) getAccessToken(ctx context.Context, channel courier.Channel, clog *courier.ChannelLog) (string, error) {
+	rc := h.Backend().RedisPool().Get()
+	defer rc.Close()
+
+	tokenKey := fmt.Sprintf("channel-token:%s", channel.UUID())
+
+	h.fetchTokenMutex.Lock()
+	defer h.fetchTokenMutex.Unlock()
```
rowanseymour, line 192: I know this is how it's implemented for the other channels like this.. but I remember thinking this is potentially a race condition across different boxes (depending on how MTN API works). 2 different courier boxes could get in here at the same time and fetch new access tokens at the same time.. 

If requesting the second one invalidates the first one.. then there's a problem. If both tokens are valid then there's no problem. Not a huge problem either way because even if the first token is now invalid, we fail to send that mesage, but we'll get the second token when we try again.

All that to say we can merge it like this, but we should look into this.

rowanseymour (approved):

PR #576, opened 2023-03-20
Add way to customize the API host for MTN channels in channel config

norkans7:


handlers/mtn/mtn.go
```diff
@@ -20,9 +20,9 @@ import (
 )
 
 var (
-	sendURL      = "https://api.mtn.com/v2/messages/sms/outbound"
-	maxMsgLength = 160
-	tokenURL     = "https://api.mtn.com/v1/oauth/access_token?grant_type=client_credentials"
+	APIHostURL    = "https://api.mtn.com"
```
norkans7, line 23: The team suggested we use a different URL for testing the messages

rowanseymour, line 23: `apiHostURL`

rowanseymour (approved):

norkans7:
@rowanseymour I am adding other changes suggested to this

handlers/mtn/mtn.go
```diff
@@ -120,46 +120,44 @@ func (h *handler) Send(ctx context.Context, msg courier.Msg, clog *courier.Chann
...
-			return status, nil
-		}
-
-		// if this is our first message, record the external id
-		if i == 0 {
-			status.SetExternalID(externalID)
-			status.SetStatus(courier.MsgWired)
-		}
 
+	mtMsg := &mtPayload{}
```
norkans7, line 128: we do not need to divide the message in parts, they do

norkans7: and since we need the clientCorrelator to be unique for each message, I used the UUID

I am not sure how our retry mechanism  in case of error will work but we'll see

rowanseymour: So `clientCorrelator` is like an external id except we can generate it? `Msg.UUID` isn't indexed so we don't want to lookup msgs by that. But we could put msg.id in `clientCorrelator` and just use that for lookups.

rowanseymour (approved):

PR #594, opened 2023-06-14
Reduce repeated code

norkans7:
When adding D3C I noticed there is a lot a codes repeated between D3C and WAC, this is trying to reduce the repeated code and make that more DRY


norkans7:
This now needs to be updated first because of #602

rowanseymour:
`utils` doesn't feel like the right place for shared WhatsApp functionality between handlers, and this PR is creating functions like `utils.GetSupportedLanguages` which are WhatsApp specific but it's certainly not obvious from the name. And then we also have structs like `handlers.MsgTemplating` which are WhatsApp specific but again not obvious from the name.

Thus far `courier.*` has been the high level interface, `handlers.*` has been support functionality for building handlers, and `utils.*` is some very generic HTTP stuff.

I think my preference for stuff which is WhatsApp specific but shared between multiple handlers, is put it in `handlers/whatsapp.go` but prefix it with `WA` or `WAC`

norkans7:
I think I had one issue of circular dependencies but I will try that again and see why I had added that in utils

rowanseymour:
Closing in favor of the direction started in https://github.com/nyaruka/courier/pull/631

PR #596, opened 2023-06-30
Log commiter using batcher

rowanseymour:
Rework log committer to use new generic Batcher type which will be easier to rework to support S3 logs as well.

Will convert status committer too in subsequent PR.


backends/rapidpro/channel_log.go
```diff
@@ -25,11 +31,6 @@ type ChannelLog struct {
 	ElapsedMS int                    `db:"elapsed_ms"`
 }
 
-// RowID satisfies our batch.Value interface, we are always inserting logs so we have no row id
-func (l *ChannelLog) RowID() string {
```
rowanseymour, line 29: this is for deduping which isn't needed for inserts

PR #597, opened 2023-07-03
Rework writing msg statuses to use Batcher like channel logs

rowanseymour:


backends/rapidpro/status.go
```diff
@@ -281,6 +286,14 @@ UPDATE msgs_msg SET
...
 			next_attempt 
 		END,
+	failed_reason = CASE
+		WHEN
+			error_count >= 2
+		THEN
+			'E'
+		ELSE
+			failed_reason
+	    END,
```
rowanseymour, line 296: we have 3 different sql queries for updating message statuses and we don't do a great job of keeping them all in sync.. gonna look at reducing that to two queries

PR #600, opened 2023-07-12
Add `log_policy` field to channel

rowanseymour:
Which lets us selectively discard channel logs.

Will require changes on RP side to add and backfill that field.

### DIFF
--- a/backends/rapidpro/channel.go
+++ b/backends/rapidpro/channel.go
@@ -15,6 +15,14 @@ import (
 	"github.com/nyaruka/null/v2"
 )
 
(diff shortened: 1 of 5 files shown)


backends/rapidpro/channel.go
```diff
@@ -15,6 +15,14 @@ import (
 	"github.com/nyaruka/null/v2"
 )
 
+type LogPolicy string
+
+const (
+	LogPolicyAll    = "all"
```
norkans7, line 21: Should we use `CharField(max_length=1)`?

PR #604, opened 2023-07-26
Support caching msg external ID to match that easily even before the …

norkans7:
…status are committed to the DB

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -288,6 +288,16 @@ func (b *backend) NewMsgStatusForID(channel courier.Channel, id courier.MsgID, s
 
 // NewStatusUpdateForID creates a new Status object for the given message id
 func (b *backend) NewMsgStatusForExternalID(channel courier.Channel, externalID string, status courier.MsgStatusValue, clog *courier.ChannelLog) courier.MsgStatus {
+	rc := b.RedisPool().Get()
+	defer rc.Close()
+
+	key := fmt.Sprintf("msg-external-id|%s|%s", string(channel.UUID()), externalID)
+
+	msgID, err := redis.Int64(rc.Do("GET", key))
+	if err == nil {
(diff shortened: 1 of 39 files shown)


rowanseymour:
Two problems with this approach:

 1. We don't want to use a redis key per msg as that creates performance issues for redis to manage and expire so many keys. We should use `redisx.IntervalHash` like the other cached things in the backend.
 2. We want this to be transparent to handlers. I.e. they just ask the backend to write an external id based status update, the backend converts that to an id based one if possible.

Something else to think about is ordering across different courier instances on different boxes. If box 1 queues a status update like `id=123 ext=EX432 status=S` and box 2 a status update like `ext=EX432 status=D`... how do we ensure those result in a message with `ext=EX432` and `status=D`.

norkans7:
On 2, I think the problem we are trying to address is to be able to match the correct msg when the status has not yet been written to the DB.
So we add the ext_id mapping to the msg_id and 
the backend is trying to lookup the msg_id for the external ID exists in the redis 
https://github.com/nyaruka/courier/pull/604/files#diff-cd80dd94c961de548e47fa3ccde4e9abe56a2512db695b37b41822bf247ea31eR291-R300

If so use the msg ID, other wise keep the usual way.

Do you mean we pass a redis pool and the msg to `MsgStatus.SetExternalID` instead of using the handlers helper?
That is how I had started but I thought `MsgStatus` did not need to know about the redis and msg

rowanseymour:
I was thinking `Backend.WriteMsgStatus` could do the conversion.. we could make it so that updates are always id based - if the id isn't in redis we just look it up from the db. Assuming that ids generally will be in redis (have an interval hash that lasts 24 hours) then we shouldn't generally end up hitting the database ever.

Currently there are 3 different ways a message status can be updated and 3 different mega queries which we haven't done a good job of keeping in sync. I wanna reduce that to just 1... I'm going to do a little refactoring now so maybe don't start changing anything...

PR #609, opened 2023-07-27
Rework deduping of incoming messages

rowanseymour:
Merges deduping by text and by ext id - if we have an ext id, we don't use text.

### DIFF
--- a/backends/rapidpro/backend_test.go
+++ b/backends/rapidpro/backend_test.go
@@ -737,36 +737,36 @@ func (ts *BackendTestSuite) TestHeartbeat() {
 	ts.NoError(ts.b.Heartbeat())
 }
 
-func (ts *BackendTestSuite) TestDupes() {
+func (ts *BackendTestSuite) TestCheckForDuplicate() {
 	r := ts.b.redisPool.Get()
 	defer r.Close()
 
 	ctx := context.Background()
 	knChannel := ts.getChannel("KN", "dbc126ed-66bc-4e28-b67b-81dc3327c95d")
-	clog := courier.NewChannelLog(courier.ChannelLogTypeUnknown, knChannel, nil)
+	twChannel := ts.getChannel("TW", "dbc126ed-66bc-4e28-b67b-81dc3327c96a")
 	urn, _ := urns.NewTelURNForCountry("12065551215", knChannel.Country())
+	urn2, _ := urns.NewTelURNForCountry("12065551277", knChannel.Country())
 
-	msg := ts.b.NewIncomingMsg(knChannel, urn, "ping", clog).(*DBMsg)
-	err := ts.b.WriteMsg(ctx, msg, clog)
-	ts.NoError(err)
+	createAndWriteMsg := func(ch courier.Channel, u urns.URN, text, extID string) *DBMsg {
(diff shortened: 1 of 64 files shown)


handlers/dialog360/dialog360.go
```diff
@@ -283,8 +283,7 @@ func (h *handler) processCloudWhatsAppPayload(ctx context.Context, channel couri
 				}
 
 				// create our message
-				ev := h.Backend().NewIncomingMsg(channel, urn, text, clog).WithReceivedOn(date).WithExternalID(msg.ID).WithContactName(contactNames[msg.From])
-				event := h.Backend().CheckExternalIDSeen(ev)
```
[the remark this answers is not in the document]
rowanseymour: Added that in 57c1ea15654341fd08be489ae7a8f37222a06689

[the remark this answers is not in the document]
norkans7: This was to mainly dedupe by external ID, we do that in the loop since whatsapp can send multiple message in one request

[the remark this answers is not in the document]
norkans7: In case on message is sent multiple times in different requests, that was not concerned by duplicates in the same request, but having that is probably nicer too

backends/rapidpro/msg.go
```diff
@@ -262,63 +265,61 @@ func (b *backend) flushMsgFile(filename string, contents []byte) error {
...
 
-// clearMsgSeen clears our seen incoming messages for the passed in channel and URN
-func (b *backend) clearMsgSeen(rc redis.Conn, msg *DBMsg) {
-	b.seenMsgs.Remove(rc, msg.fingerprint(false))
+			// if it is the same, return the UUID
+			if prevText == msg.Text() {
+				return courier.MsgUUID(uuidAndText[:36])
+			}
+		}
+	}
```
norkans7, line 297: I understand the issue was about Telegram when we started this that gives us external IDs, but about making sure that any channel type that will send 2 messages without text but different attachments are also save as too messages, I think here, they will not as long as the are sent with the same timestamp

backends/rapidpro/backend_test.go
```diff
@@ -777,39 +777,33 @@ func (ts *BackendTestSuite) TestDupes() {
...
-	err = ts.b.WriteMsg(ctx, msg, clog)
-	ts.NoError(err)
-
-	ts.NotEqual(uuid2, msg.UUID())
-}
-
-func (ts *BackendTestSuite) TestExternalIDDupes() {
-	r := ts.b.redisPool.Get()
-	defer r.Close()
+	msg4 := createAndWriteMsg(knChannel, urn, "test", "")
```
norkans7, line 785: Can we also have test of a message with the same text but now with an attachment or 2 messages with the same text but different attachments? without external IDs?

That is the cases I think we miss only from these changes

backends/rapidpro/msg.go
rowanseymour: Yeah tho I can't imagine there are many channels that send attachments but don't give us external ids.

We can merge this and then you can add your previous changes to use a hash of text+attachments.

norkans7: Right, that is not something that is likely to happen it is just that I had found that is could be an issue in theory 

I can add support for that after the merge

norkans7 (approved):

PR #610, opened 2023-07-28
Support sending and receiving stickers for WAC and D3C

norkans7:
This adds support to send and receive stickers as documented on https://developers.facebook.com/docs/whatsapp/cloud-api/reference/media#supported-media-types

The file need to be for format `image/webp` and size less than 100KB or 500KB for animated one with dimensions not exceeding 512x512


rowanseymour:
@norkans7 since this isn't a bug gonna hold off merging until after release

handlers/dialog360/dialog360.go
```diff
@@ -621,7 +632,9 @@ func (h *handler) Send(ctx context.Context, msg courier.Msg, clog *courier.Chann
 
 		} else if i < len(msg.Attachments()) && (len(qrs) == 0 || len(qrs) > 3) {
 			attType, attURL := handlers.SplitAttachment(msg.Attachments()[i])
-			attType = strings.Split(attType, "/")[0]
+			splitedAttType := strings.Split(attType, "/")
```
rowanseymour, line 635: @norkans7 how about we rework this to call resolve media and then we can also check that the image is 512 x 512 ?

norkans7: Okay, I'll update that

norkans7:
contains #624

norkans7:
This branch is out of date I will rework on that again

PR #614, opened 2023-08-16
Add support for receiving MO JSON body

norkans7:
So far we support receiving Form and XML only

For aggregator that forward message as JSON body only we can support them with this PR changes


rowanseymour:
If we don't need this now, I'm inclined to not make that channel type any more complicated

PR #616, opened 2023-08-24
Update channel type to save external ID for MO messages if we can, so…

norkans7:
… we can dedupe by that


handlers/dart/dart.go
```diff
@@ -56,9 +56,10 @@ func (h *handler) Initialize(s courier.Server) error {
 }
 
 type moForm struct {
-	Message  string `name:"message"`
-	Original string `name:"original"`
-	SendTo   string `name:"sendto"`
+	Message    string `name:"message"`
+	Original   string `name:"original"`
+	SendTo     string `name:"sendto"`
+	ExternalID string `name:"messageid"`
```
rowanseymour, line 62: feels confusing to not have the field name match the the JSON

rowanseymour (changes requested):

rowanseymour (approved):

PR #617, opened 2023-08-29
Rework writing status updates so that updates by external id also use the batcher

rowanseymour:
This resolves external ids to database ids in the batcher using a database lookup - I think it makes sense to optimize that by stuffing external ids in redis, but also it's not going to be much of a performance hit to not do that for now since the lookup is in batches - and I've checked it's just an index scan on `msgs_by_external_id`.

Of course the other reason to put external ids in redis is to avoid the race condition of trying to do a status update by external id before the status update that sets the external id. This won't fix that but it should make it less likely because previously the status update to write the external id was being done async (always a delay) but a subsequent update using that external id would by sync (no delay)... now they're both queued, both async, both with a delay.

### DIFF
--- a/backends/rapidpro/backend_test.go
+++ b/backends/rapidpro/backend_test.go
@@ -606,19 +606,9 @@ func (ts *BackendTestSuite) TestMsgStatus() {
 	ts.Equal(courier.MsgDelivered, m.Status_)
 	ts.NotNil(m.SentOn_)
 
-	// no such external id for outgoing message
-	status := ts.b.NewMsgStatusForExternalID(channel, "ext2", courier.MsgSent, clog6)
-	err := ts.b.WriteMsgStatus(ctx, status)
-	ts.Error(err)
-
-	// no such external id
-	status = ts.b.NewMsgStatusForExternalID(channel, "ext3", courier.MsgSent, clog6)
-	err = ts.b.WriteMsgStatus(ctx, status)
-	ts.Error(err)
-
 	// reset our status to sent
-	status = ts.b.NewMsgStatusForExternalID(channel, "ext1", courier.MsgSent, clog6)
-	err = ts.b.WriteMsgStatus(ctx, status)
+	status := ts.b.NewMsgStatusForExternalID(channel, "ext1", courier.MsgSent, clog6)
+	err := ts.b.WriteMsgStatus(ctx, status)
 	ts.NoError(err)
 	time.Sleep(time.Second)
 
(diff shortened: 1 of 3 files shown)


backends/rapidpro/backend_test.go
```diff
@@ -606,19 +606,9 @@ func (ts *BackendTestSuite) TestMsgStatus() {
 	ts.Equal(courier.MsgDelivered, m.Status_)
 	ts.NotNil(m.SentOn_)
 
-	// no such external id for outgoing message
-	status := ts.b.NewMsgStatusForExternalID(channel, "ext2", courier.MsgSent, clog6)
-	err := ts.b.WriteMsgStatus(ctx, status)
```
rowanseymour, line 611: finding out that an external id doesn't exist is now async

backends/rapidpro/status.go
```diff
@@ -341,3 +350,42 @@ func writeMsgStatuses(ctx context.Context, db *sqlx.DB, spoolDir string, statuse
 		}
 	}
 }
+
+const sqlResolveStatusMsgIDs = `
+SELECT id, channel_id, external_id 
+  FROM msgs_msg 
+ WHERE (channel_id, external_id) IN (VALUES(CAST(:channel_id AS int), :external_id))`
```
rowanseymour, line 357: note that you can't write `:channel_id::int` because sqlx will get confused booo

backends/rapidpro/status.go
```diff
@@ -341,3 +350,42 @@ func writeMsgStatuses(ctx context.Context, db *sqlx.DB, spoolDir string, statuse
...
+	var channelID courier.ChannelID
+	var externalID string
+
+	for rows.Next() {
+		if err := rows.Scan(&msgID, &channelID, &externalID); err != nil {
+			return errors.Wrap(err, "error scanning rows")
+		}
+
+		// find the status with this channel ID and external ID and update its msg ID
+		for _, s := range statuses {
```
norkans7, line 383: Can we build a map here and loop on the statuses outside the for loop for rows?

rowanseymour: I guess so.. I went back and forth on that because the map has to use a compound key of channel + external id.. and I persuaded myself that the complexity of creating that isn't worth the performance gains.. but lemme try it.

rowanseymour: Ok it isn't too bad.. updated

norkans7 (approved):

PR #618, opened 2023-08-30
Rework writing msg statuses to always use id resolving

rowanseymour:
* Ensure that if we get an error trying to resolve a msg id, that msg is spooled
* Change flushing spooled statuses to use the same code path as the status writer

### DIFF
--- a/backends/rapidpro/backend_test.go
+++ b/backends/rapidpro/backend_test.go
@@ -2,7 +2,6 @@ package rapidpro
 
 import (
 	"context"
-	"database/sql"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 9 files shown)


backends/rapidpro/backend_test.go
```diff
@@ -169,36 +168,6 @@ func (ts *BackendTestSuite) TestMsgUnmarshal() {
 	ts.Equal("", msg.FlowUUID())
 }
 
-func (ts *BackendTestSuite) TestCheckMsgExists() {
```
rowanseymour, line 172: these are tests.. for a method only used in tests...

backends/rapidpro/status.go
```diff
@@ -106,114 +105,18 @@ WHERE
...
-	if err == courier.ErrMsgNotFound {
-		return nil
-	}
-
-	// Ignore wrong status update for incoming messages
-	if err == courier.ErrWrongIncomingMsgStatus {
-		return nil
-	}
-
+	_, err = writeMsgStatusesToDB(ctx, b.db, []*DBMsgStatus{status})
```
rowanseymour, line 119: if the message can't be resolved/found.. this call doesn't error now

backends/rapidpro/status.go
```diff
@@ -106,114 +105,18 @@ WHERE
...
 	// try to flush to our db
-	err = writeMsgStatusToDB(context.Background(), b, status)
-
-	// not finding the message is ok for status updates
-	if err == courier.ErrMsgNotFound {
-		return nil
-	}
-
-	// Ignore wrong status update for incoming messages
-	if err == courier.ErrWrongIncomingMsgStatus {
```
rowanseymour, line 213: `ErrWrongIncomingMsgStatus` isn't be returned by anything

handlers/dialog360/dialog360.go
```diff
@@ -328,13 +328,6 @@ func (h *handler) processCloudWhatsAppPayload(ctx context.Context, channel couri
 
 				event := h.Backend().NewMsgStatusForExternalID(channel, status.ID, msgStatus, clog)
 				err := h.Backend().WriteMsgStatus(ctx, event)
-
-				// we don't know about this message, just tell them we ignored it
-				if err == courier.ErrMsgNotFound {
-					data = append(data, courier.NewInfoData(fmt.Sprintf("message id: %s not found, ignored", status.ID)))
-					continue
```
rowanseymour, line 335: all message status writes are async now so this can never happen

norkans7 (approved):

PR #620, opened 2023-08-31
Update to new batchers in gocommon

rowanseymour:


backends/rapidpro/channel_log.go
```diff
@@ -120,28 +119,26 @@ func NewDBLogWriter(db *sqlx.DB, wg *sync.WaitGroup) *DBLogWriter {
 			defer cancel()
 
 			writeDBChannelLogs(ctx, db, batch)
-		}, time.Millisecond*500, 1000, wg),
+		}, 1000, time.Millisecond*500, 1000, wg),
 	}
 }
 
-func writeDBChannelLogs(ctx context.Context, db *sqlx.DB, logs []*dbChannelLog) {
-	for _, batch := range utils.ChunkSlice(logs, 1000) {
```
rowanseymour, line 128: no longer need to batch here since the batcher will only ever return batches <= 1000

backends/rapidpro/backend_test.go
```diff
@@ -473,7 +473,7 @@ func (ts *BackendTestSuite) TestMsgStatus() {
 		}
 		err := ts.b.WriteStatusUpdate(ctx, statusObj)
 		ts.NoError(err)
-		time.Sleep(500 * time.Millisecond) // give committer time to write this
```
rowanseymour, line 476: previously the timeout started when the batcher was created.. now it starts whenever a value is queued so we have to give the test a little more leeway to be sure the batcher triggers a batch

PR #623, opened 2023-08-31
Record external ids of sent messages in redis

rowanseymour:
and use for resolving ids for status updates

### DIFF
--- a/backends/rapidpro/backend.go
+++ b/backends/rapidpro/backend.go
@@ -96,10 +96,9 @@ func newBackend(cfg *courier.Config) courier.Backend {
 		mediaCache:   redisx.NewIntervalHash("media-lookups", time.Hour*24, 2),
 		mediaMutexes: *syncx.NewHashMutex(8),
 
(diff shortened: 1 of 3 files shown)


backends/rapidpro/backend.go
```diff
@@ -536,11 +542,24 @@ func (b *backend) WriteStatusUpdate(ctx context.Context, status courier.StatusUp
...
-		err := b.ClearMsgSent(ctx, status.ID())
-		if err != nil {
-			logrus.WithError(err).WithField("msg", status.ID()).Error("error clearing sent flags")
+	if status.ID() != courier.NilMsgID {
+		// this is a message we've just sent and were given an external id for
+		if status.ExternalID() != "" {
+			rc := b.redisPool.Get()
+			defer rc.Close()
+
+			err := b.sentExternalIDs.Set(rc, fmt.Sprintf("%d|%s", su.ChannelID_, su.ExternalID_), fmt.Sprintf("%d", status.ID()))
```
rowanseymour, line 551: we should update the other hashes to use channel id instead of uuid.. there's no advantage to using uuid here and it's more redis storage

norkans7 (approved):

PR #625, opened 2023-09-05
Rework handler tests so that test cases must explicitly say if they don't generate a channel log

rowanseymour:


handlers/bongolive/bongolive_test.go
```diff
@@ -46,8 +46,8 @@ var testCases = []ChannelHandleTestCase{
 	{
 		Label:                "Status No params",
 		URL:                  receiveURL,
-		Data:                 "",
-		ExpectedRespStatus:   405,
+		Data:                 "&",
```
rowanseymour, line 49: this wasn't actually testing a POST with no params because Data = "" means use GET

handlers/telegram/telegram_test.go
```diff
@@ -693,7 +694,7 @@ var testCases = []ChannelHandleTestCase{
 		Data:                 errorFile,
 		ExpectedRespStatus:   200,
 		ExpectedBodyContains: "unable to resolve file",
-		ExpectedErrors:       []*courier.ChannelError{courier.ErrorResponseUnparseable("JSON")},
+		ExpectedErrors:       []*courier.ChannelError{courier.ErrorExternal("500", "error loading file")},
```
rowanseymour, line 697: this wasn't being checked because this we previously required a test case to return a message or status before we check channel logs

PR #626, opened 2023-09-05
Don't rely on knowing msg id to determine if a log is attached

rowanseymour:


server.go
```diff
@@ -331,11 +331,11 @@ func (s *server) channelHandleWrapper(handler ChannelHandler, handlerFunc Channe
 			for _, event := range events {
 				switch e := event.(type) {
 				case Msg:
-					clog.SetMsgID(e.ID())
```
rowanseymour, line 334: problem is we don't know msg id for most status updates because they're by external id

PR #627, opened 2023-09-06
No need to try making DB queries when all msg IDs got resolved from r…

norkans7:
…edis

Fixes https://textit.sentry.io/issues/4457890102/

### DIFF
--- a/backends/rapidpro/backend_test.go
+++ b/backends/rapidpro/backend_test.go
@@ -700,6 +700,9 @@ func (ts *BackendTestSuite) TestSentExternalIDCaching() {
 	err := ts.b.WriteStatusUpdate(ctx, status1)
 	ts.NoError(err)
 
+	// give batcher time to write it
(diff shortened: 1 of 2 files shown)


backends/rapidpro/backend_test.go
```diff
@@ -708,6 +708,9 @@ func (ts *BackendTestSuite) TestSentExternalIDCaching() {
 	// mimic a delay in that status being written by reverting the db changes
 	ts.b.db.MustExec(`UPDATE msgs_msg SET status = 'W', external_id = NULL WHERE id = 10000`)
 
+	// Add 1 second delay to make sure the status update have distict times
+	time.Sleep(time.Millisecond * 1000)
```
rowanseymour, line 712: I think move this line up to 703... this is so that the status update is written by the batcher no? And make it 600 ms

rowanseymour (approved):

PR #629, opened 2023-09-06
Update to latest null library and use Map[string] for channel event extra

rowanseymour:


backends/rapidpro/backend_test.go
```diff
@@ -1269,7 +1269,7 @@ func (ts *BackendTestSuite) TestChannelEvent() {
 	clog := courier.NewChannelLog(courier.ChannelLogTypeUnknown, channel, nil)
 	urn, _ := urns.NewTelURNForCountry("12065551616", channel.Country())
 
-	event := ts.b.NewChannelEvent(channel, courier.Referral, urn, clog).WithExtra(map[string]any{"ref_id": "12345"}).WithContactName("kermit frog")
```
rowanseymour, line 1272: it's easy to make this less strict if we need in future.. but if we don't need, then it's nice to know we can just read values as strings

PR #636, opened 2023-09-12
Allow handlers to create arbitrary auth tokens with messages and channel events

rowanseymour:


test/msg.go
```diff
@@ -64,31 +65,39 @@ func (m *mockMsg) FlowUUID() string {
...
+func (m *mockMsg) ID() courier.MsgID                { return m.id }
+func (m *mockMsg) EventID() int64                   { return int64(m.id) }
+func (m *mockMsg) UUID() courier.MsgUUID            { return m.uuid }
+func (m *mockMsg) Text() string                     { return m.text }
+func (m *mockMsg) Attachments() []string            { return m.attachments }
+func (m *mockMsg) Locale() i18n.Locale              { return m.locale }
+func (m *mockMsg) ExternalID() string               { return m.externalID }
+func (m *mockMsg) URN() urns.URN                    { return m.urn }
+func (m *mockMsg) URNAuth() string                  { return m.urnAuth }
+func (m *mockMsg) URNAuthTokens() map[string]string { return m.urnAuthTokens }
```
rowanseymour, line 78: this is bit confusing but I think we should move methods that are only used for testing elsewhere

norkans7 (approved):

PR #637, opened 2023-09-13
Simpler `Msg` interface

rowanseymour:
Methods only used for testing only need to be on `MockMsg`

### DIFF
--- a/msg.go
+++ b/msg.go
@@ -78,20 +78,9 @@ type Msg interface {
 	HighPriority() bool
 
 	// incoming specific
-	URNAuthTokens() map[string]string
-	ContactName() string
 	ReceivedOn() *time.Time
-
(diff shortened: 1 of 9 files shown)


msg.go
```diff
@@ -52,46 +52,35 @@ const (
...
 	Attachments() []string
-	Locale() i18n.Locale
-	ExternalID() string
 	URN() urns.URN
-	URNAuth() string
-	URNAuthTokens() map[string]string
-	ContactName() string
+	Channel() Channel
+
+	// outgoing specific
```
rowanseymour, line 65: might be even cleaner to have separate interfaces for in and out...

norkans7: I will try to make that change

norkans7 (approved):

PR #639, opened 2023-09-13
Add support for FB notification messages optin and optout events

rowanseymour:


test/msg.go
```diff
@@ -84,10 +84,6 @@ func (m *MockMsg) WithURNAuthTokens(tokens map[string]string) courier.Msg {
 }
 func (m *MockMsg) WithReceivedOn(date time.Time) courier.Msg { m.receivedOn = &date; return m }
 
-// used for testing created incoming messages
-func (m *MockMsg) URNAuthTokens() map[string]string { return m.urnAuthTokens }
-func (m *MockMsg) ContactName() string              { return m.contactName }
```
rowanseymour, line 89: don't need these because we're checking the values stored on the mock backend instead

utils/whatsapp/api.go
```diff
@@ -136,6 +136,16 @@ type MOPayload struct {
 			Timestamp int64 `json:"timestamp"`
 
 			OptIn *struct {
```
rowanseymour, line 138: @norkans7 I didn't realize whatsapp and facebook were using the same structs here... should they be? Should this package be the `meta` package instead of the `whatsapp` package?

channel_event.go
```diff
@@ -15,6 +15,8 @@ const (
 	Referral        ChannelEventType = "referral"
 	StopContact     ChannelEventType = "stop_contact"
 	WelcomeMessage  ChannelEventType = "welcome_message"
+	EventTypeOptIn  ChannelEventType = "optin"
+	EventTypeOptOut ChannelEventType = "optout"
```
rowanseymour, line 19: keeping these very generic and not FB specific

utils/whatsapp/api.go
norkans7: yes, they share the structs. we can rename that package

norkans7 (approved):

PR #642, opened 2023-09-15
Meta cleanup

rowanseymour:


handlers/whatsapp_legacy/handler.go
```diff
@@ -1147,3 +1147,96 @@ type MsgTemplating struct {
 	Namespace string   `json:"namespace"`
 	Variables []string `json:"variables"`
 }
+
+func getSupportedLanguage(lc i18n.Locale) string {
```
rowanseymour, line 1151: this is just putting back the previous code so that this package doesn't need to depend on anything in the `meta` package

PR #644, opened 2023-09-18
Msg interfaces separation

norkans7:
This separates the `courier.Msg` interfaces into 2 interfaces `courier.MsgOut` for outgoing messages and `courier.MsgIn` for incoming messages.
And changes all methods to expect the proper interface according to whether they are to handle for outgoing or incoming messages.


rowanseymour (approved):
Nice! Feels a lot cleaner

PR #660, opened 2023-10-12
Finish switching to use slog for logging 

norkans7:
Removes logrus use and add sentry handler for slog


cmd/courier/main.go
```diff
@@ -87,51 +89,61 @@ func main() {
...
-			log.Error("unable to configure sentry hook", "dsn", config.SentryDSN, "error", err)
+			log.Fatalf("error initiating sentry client, error %s, dsn %s", err, config.SentryDSN)
 			os.Exit(1)
 		}
-		logrus.StandardLogger().Hooks.Add(hook)
+
+		defer sentry.Flush(2 * time.Second)
+
+		logger = slog.New(
+			slogmulti.Fanout(
```
norkans7, line 121: We need to use multi handlers here so that we can log the error both to sentry and the text logs on the servers

cmd/courier/main.go
```diff
@@ -87,51 +89,61 @@ func main() {
 	}
 
 	// configure our logger
-	logrus.SetOutput(os.Stdout)
-	level, err := logrus.ParseLevel(config.LogLevel)
+	loggerLevel := new(slog.LevelVar)
```
rowanseymour, line 92: do we need this? or can we just parse the `Level` value and then create the handler?

norkans7: using the levelVar is the only approach to support later changing the log level, but we change to use the parsed level and create a new handler.

rowanseymour (approved):
Woohoo!

PR #662, opened 2023-11-29
Prep for support for more WA templates components

norkans7:
needs adjustment to the engine to send the the components in metadata instead of variables


handlers/whatsapp_legacy/handler_test.go
```diff
@@ -1070,21 +1070,6 @@ var mediaCacheSendTestCases = []OutgoingTestCase{
 	},
 }
 
-var hsmSupportSendTestCases = []OutgoingTestCase{
```
norkans7, line 1073: hsm type was replaced by template, so removing support for that as that is no longer used

PR #666, opened 2023-12-04
Fix FBA timestamps that sometimes are in seconds instead of milliseconds

norkans7:


handlers/meta/handlers.go
```diff
@@ -411,7 +411,7 @@ func (h *handler) processFacebookInstagramPayload(ctx context.Context, channel c
 
 		var date time.Time
 
-		if msg.Timestamp >= 10_000_000_000 {
+		if msg.Timestamp >= 100_000_000_000 {
```
norkans7, line 414: @rowanseymour I found the right number to use is 100_000_000_000 that we cannot consider as milliseconds
after experimenting with https://www.epochconverter.com/

PR #670, opened 2023-12-12
Make kannel channel paused when we get Queued message in response

norkans7:


handlers/kannel/handler.go
```diff
@@ -200,10 +201,22 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, clog *courier.Ch
...
+	}
+
+	if strings.Contains(string(respBody), "Queued") {
+		rc := h.Backend().RedisPool().Get()
+		defer rc.Close()
+		rateLimitKey := fmt.Sprintf("rate_limit:%s", msg.Channel().UUID())
+		rc.Do("SET", rateLimitKey, "engaged")
+
+		// We pause sending 30 seconds so the connection to the SMSC is reset
+		rc.Do("EXPIRE", rateLimitKey, 30)
```
rowanseymour, line 218: SET + EXPIRE = SETNX.. no ?

norkans7: That is deprecated https://redis.io/commands/setnx/

But I can do on line as `rc.Do("SET", rateLimitKey, "engaged", "EX", 30)`

handlers/kannel/handler.go
```diff
@@ -200,10 +201,20 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, clog *courier.Ch
...
+		resp, respBody, err = h.RequestHTTP(req, clog)
 	} else {
-		resp, _, err = h.RequestHTTPInsecure(req, clog)
+		resp, respBody, err = h.RequestHTTPInsecure(req, clog)
+	}
+
+	if strings.Contains(string(respBody), "Queued") {
+		rc := h.Backend().RedisPool().Get()
+		defer rc.Close()
+		rateLimitKey := fmt.Sprintf("rate_limit:%s", msg.Channel().UUID())
```
rowanseymour, line 214: It doesn't feel ideal having handlers talk directly to Redis to rate limit themselves and I see we only currently do this in the old WhatsApp onprem handler. And there's `rate_limit_bulk`.. don't we need to rate limit on both queues?

Maybe we need to have a think about making a cleaner way for channels to rate limit themselves?

norkans7: Okay, we can find a new way tor that.

I do not think we need to pause the bulk queue in this case but I will double check on that again

norkans7: Added pausing the bulk queue as well

PR #673, opened 2024-01-10
Temba Chat

rowanseymour:


handlers/tembachat/handler_test.go
```diff
@@ -0,0 +1,77 @@
...
+		ExpectedBodyContains: "Field validation for 'Type' failed on the 'required' tag",
+	},
+}
+
+func TestIncoming(t *testing.T) {
+	RunIncomingTestCases(t, testChannels, newHandler(), handleTestCases)
+}
+
+func setSendURL(s *httptest.Server, h courier.ChannelHandler, c courier.Channel, m courier.MsgOut) {
+	defaultSendURL = s.URL
```
norkans7, line 47: I think it i better to set the config send URL here and remove `defaultSendURL`

rowanseymour: I'm just copying how other handlers are done but I did wonder if all handlers should support `send_url` in config and then this can be the normal way of overriding it in tests

norkans7: that could work too

PR #681, opened 2024-01-22
Add user email and name to queued messages

rowanseymour:
Just using user id worked with temba-chat because it can look up users, but we also want the new mailgun handler to know about who sent a message.. without courier having to look up users from the database.


ericnewcomer:
...its prob going to want to know avatar url though.

rowanseymour:
I think that's on temba-chat to lookup... don't want courier queues to end up full of stuff most channels don't use.

ericnewcomer:
You don't think the email channel wants this?

rowanseymour:
Hmm you mean like rich html emails with avatars ? That's a tricky one.. would be nice if there was a way for RP to generate those using templates etc.

PR #684, opened 2024-02-02
Support MarkdownV2 text formatting styles

norkans7:
Add support for formatting options
https://core.telegram.org/bots/api#formatting-options

As that is not exactly how other channels such as WhatsApp do that https://faq.whatsapp.com/539178204879377/

I think the user will be responsible to make sure the text will be formatted properly for each channel type


rowanseymour:
I just tested this an my flows are broken because they use `!`... I think I need to revert this.

We want as much as possible to allow users to create the same messages across different channels. Maybe there's a way to let users format with the very basic `_hello_` `*hello*` syntax (that conveniently WhatsApp uses).. and convert that where needed in channel handlers. @ericnewcomer

norkans7:
i had though the user will be splitting by channel when they want to use that formatting

norkans7:
Or then support the old Markdown style only
https://core.telegram.org/bots/api#markdown-style

Issue #653, opened 2023-10-04
Improve send retry logic


rowanseymour:
It made sense back in the day when we were mostly hitting unreliable SMS providers to be liberal about retries but it makes no sense to do 3 retries when a channel type like Facebook gives us a clear error code. The only things we should be retrying are probably:

1. Connection timeout
2. Rate limited 429 responses (but even here we might want to be smarter)
3. 500 responses

rowanseymour:
Pondering on whether it would make `Send` methods generally simpler if they didn't have to construct a `StatusUpdate` but instead took an interface as a param with methods like..

`SetFailed`
`SetExternalID`

which is then turned into a regular `StatusUpdate`.. but maybe if a `Send` returns an error that means queue up a `StatusUpdate` with status=ERRORED (retryable)

rowanseymour:
Ok bit more pondering on how to make the contract for sending simpler for handlers, so our logic for what gets retried can be consistent and not require additional complexity within handlers..

```go
type SendResult interface {
    AddExternalID(string)
    SetFailed()
}

type Handler interface {
    ...
    Send(context.Context, Msg, SendResult, ChannelLog) error 
}
```

1. All outgoing requests succeed
   - channel log records successful requests
   - handler `Send` returns `nil` (implicit send success)
   - sender queues `status=wired` status update
2. An outgoing request times out or gets 5XX response
   - channel log records failed request
   - handler `Send` returns err
   - sender logs err but it's not user facing unlike the channel logs
   - sender queues retryable `status=errored` update
3. An outgoing request fails with 4XX status
   - channel log records failed request
   - channel log records user facing error message
   - handler records send failure (`SetFailed`)
   - handler `Send` returns `nil`
   - sender queues non-retryable `status=failed` update

`AddExternalID` instead of `SetExternalID` is intentional because one database message can be sent as multiple channel messages and thus generate multiple external ids... and one day we might want to model that in the message table so that we don't get so many status updates for messages we don't know about.

@norkans7 @ericnewcomer

rowanseymour:
I think handler finding channel is missing required config fields should fail permanently with a generic user facing error and /or maybe something in sentry.

rowanseymour:
Have introduced a new `Send` method signature that only returns an error and the logic for whether the message should be retried (status=E) or not (status=F) moves to the sender. 9 of 58 handlers updated so far...

PR #687, opened 2024-02-08
Refactor sending to do less retrying and let handlers be dumber

rowanseymour:
See #653 

 * Allows handlers to implement one of two interfaces that differ by send method to allow handlers to be rewritten piece meal.
 * New `Send` contract takes away need for handlers creating `StatusUpdate` objects - they return different error singletons and then sender determines whether message should be be retried.
 * Or for things like db errors.. they can return a regular `error` and that is treated as a logged, retryable internal error
 * Have send errors become channel log errors to avoid duplicating those.

### DIFF
--- a/handlers/africastalking/handler.go
+++ b/handlers/africastalking/handler.go
@@ -114,21 +114,16 @@ func (h *handler) receiveStatus(ctx context.Context, channel courier.Channel, w 
 }
 
 // Send sends the given message, logging any HTTP calls or errors
-func (h *handler) Send(ctx context.Context, msg courier.MsgOut, clog *courier.ChannelLog) (courier.StatusUpdate, error) {
+func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.SendResult, clog *courier.ChannelLog) error {
 	isSharedStr := msg.Channel().ConfigForKey(configIsShared, false)
 	isShared, _ := isSharedStr.(bool)
 
 	username := msg.Channel().StringConfigForKey(courier.ConfigUsername, "")
-	if username == "" {
-		return nil, fmt.Errorf("no username set for AT channel")
-	}
-
 	apiKey := msg.Channel().StringConfigForKey(courier.ConfigAPIKey, "")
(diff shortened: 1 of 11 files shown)


handlers/africastalking/handler.go
```diff
@@ -114,21 +114,16 @@ func (h *handler) receiveStatus(ctx context.Context, channel courier.Channel, w
...
-
 	apiKey := msg.Channel().StringConfigForKey(courier.ConfigAPIKey, "")
-	if apiKey == "" {
-		return nil, fmt.Errorf("no API key set for AT channel")
-	}
 
-	status := h.Backend().NewStatusUpdate(msg.Channel(), msg.ID(), courier.MsgStatusErrored, clog)
+	if username == "" || apiKey == "" {
+		return courier.ErrSendChannelConfig
+	}
```
rowanseymour, line 126: not so worried about making these errors less specific.. this should only happen if we've messed up channel creation in RP and we'll get this in Sentry.

handlers/africastalking/handler_test.go
```diff
@@ -141,33 +140,32 @@ var outgoingTestCases = []OutgoingTestCase{
...
-		Label:              "Error Sending",
+		Label:              "Missing status value",
 		MsgText:            "Error Message",
 		MsgURN:             "tel:+250788383383",
 		MockResponseBody:   `{ "error": "failed" }`,
 		MockResponseStatus: 401,
 		ExpectedRequests: []ExpectedRequest{
 			{Form: url.Values{"message": {`Error Message`}, "username": {"Username"}, "to": {"+250788383383"}, "from": {"2020"}}},
 		},
-		ExpectedMsgStatus: "E",
```
norkans7, line 169: Can we keep the `ExpectedMsgStatus` for these tests as that was brought back?
@rowanseymour

rowanseymour: that field will only be used for tests that use the legacy Send method type - in the new Send methods handlers don't determine the msg status - they return an error or not... and the sender chooses the right status.

PR #690, opened 2024-02-14
Yo and Zenvia new send

norkans7:


handlers/yo/handler.go
```diff
@@ -130,40 +118,31 @@ func (h *handler) SendLegacy(ctx context.Context, msg courier.MsgOut, clog *cour
 
 			req, err := http.NewRequest(http.MethodGet, sendURL.String(), nil)
 			if err != nil {
-				return nil, err
+				return courier.ErrConnectionFailed
```
rowanseymour, line 121: these errs.. that probably can't happen can be just returned as `err` and they'll be treated as an internal error

handlers/yo/handler_test.go
```diff
@@ -56,46 +58,106 @@ func setSendURL(s *httptest.Server, h courier.ChannelHandler, c courier.Channel,
...
+			"*": {
+				httpx.NewMockResponse(200, nil, []byte(`ybs_autocreate_status=OK`)),
+			},
+		},
+		ExpectedRequests: []ExpectedRequest{{Params: url.Values{"sms_content": {"Simple Message"},
+			"destinations": {"250788383383"},
+			"ybsacctno":    {"yo-username"},
+			"password":     {"yo-password"},
+			"origin":       {"2020"},
+		}}},
```
rowanseymour, line 71: if you're rewriting all the tests to use `MockResponses`, you can get rid of `SendPrep` and put the real URL instead of `"*"`

rowanseymour (changes requested):

handlers/yo/handler_test.go
norkans7: I think for URL params the mocks are expecting the entire URL including all the parameters,
I will check how to adjust that to match the URL ignoring the params

norkans7: for now I append * to the URLs

rowanseymour: makes sense.. you're already testing that the correct params were used in expected requests

handlers/zenvia/handlers_test.go
```diff
@@ -242,163 +242,214 @@ func BenchmarkHandler(b *testing.B) {
...
+		ExpectedRequests: []ExpectedRequest{{
+			Headers: map[string]string{
+				"Content-Type": "application/json",
+				"Accept":       "application/json",
+				"X-API-TOKEN":  "zv-api-token",
+			},
+			Body: `{"from":"2020","to":"250788383383","contents":[{"type":"text","text":"No External ID"}]}`,
+		}},
+		ExpectedError: courier.ErrResponseUnexpected,
+		SendPrep:      nil,
```
rowanseymour, line 437: You can remove `SendPrep`.. no need to set to `nil`

rowanseymour (approved):

PR #691, opened 2024-02-15
Use MockResponses and ExpectedRequests in WA legacy tests

norkans7:
No change to the send method as that use status update to update the URN

### DIFF
--- a/handlers/whatsapp_legacy/handler_test.go
+++ b/handlers/whatsapp_legacy/handler_test.go
@@ -3,7 +3,6 @@ package whatsapp_legacy
 import (
 	"context"
 	"encoding/json"
-	"net/http/httptest"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


handlers/whatsapp_legacy/handler_test.go
```diff
@@ -570,58 +568,82 @@ var defaultSendTestCases = []OutgoingTestCase{
...
-		Label:               "Unicode Send",
-		MsgText:             "☺",
-		MsgURN:              "whatsapp:250788123123",
-		MockResponseBody:    `{ "messages": [{"id": "157b5e14568e8"}] }`,
-		MockResponseStatus:  201,
-		ExpectedRequestBody: `{"to":"250788123123","type":"text","text":{"body":"☺"}}`,
-		ExpectedMsgStatus:   "W",
-		ExpectedExtIDs:      []string{"157b5e14568e8"},
-		SendPrep:            setSendURL,
+		Label:   "Unicode Send",
```
rowanseymour, line 573: I've been deleting these separate "Unicode Send" test cases... it only makes sense for mobile carriers who require different encodings for non-ASCII... they aren't needed for things like WhatsApp

rowanseymour (approved):

PR #694, opened 2024-02-16
TWT, TQ, TS handlers to new send

norkans7:


handlers/telesom/handler.go
```diff
@@ -114,21 +98,22 @@ func (h *handler) SendLegacy(ctx context.Context, msg courier.MsgOut, clog *cour
...
 		}
 		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
 		resp, respBody, err := h.RequestHTTP(req, clog)
-		if err != nil || resp.StatusCode/100 != 2 {
-			return status, nil
+		if err != nil || resp.StatusCode/100 == 5 {
+			return courier.ErrConnectionFailed
+		} else if resp.StatusCode/100 != 2 {
+			return courier.ErrResponseStatus
```
norkans7, line 109: I think previously I had not noticed `courier.ErrResponseStatus` and I think this is the proper way to handle these responses

handlers/telegram/handler_test.go
```diff
@@ -778,136 +779,153 @@ func TestIncoming(t *testing.T) {
...
-		Label:              "Plain Send",
-		MsgText:            "Simple Message",
-		MsgURN:             "telegram:12345",
-		MockResponseBody:   `{ "ok": true, "result": { "message_id": 133 } }`,
-		MockResponseStatus: 200,
+		Label:   "Plain Send",
+		MsgText: "Simple Message",
+		MsgURN:  "telegram:12345",
+		MockResponses: map[string][]*httpx.MockResponse{
+			"*/botauth_token/sendMessage": {
```
rowanseymour, line 786: I started using `*/...` url patterns where I was rewriting some tests, but other tests were still using `setSendURL` so the URL would end up being something like `http://localhost:<port>/...`... but if you find yourself rewriting ALL the tests for a handler so there no longer is any `setSendURL`.. then you use the real domain, e.g. `http://api.telegram.org/...`

norkans7: The problem here is that we mock the the URL for MO messages having attachments that is why I removed the domain so that mock service for fetching attachment can overwrite the `ApiURL` value

We can change how that works as well with a better mock later I think

rowanseymour: Ahh ok

rowanseymour (approved):

PR #695, opened 2024-02-16
Twiml handlers to new send

norkans7:


handlers/twiml/handlers.go
```diff
@@ -274,62 +263,52 @@ func (h *handler) SendLegacy(ctx context.Context, msg courier.MsgOut, clog *cour
...
+
+			return courier.ErrResponseStatus
 		}
 
 		// grab the external id
 		externalID, err := jsonparser.GetString(respBody, "sid")
 		if err != nil {
 			clog.Error(courier.ErrorResponseValueMissing("sid"))
-			return status, nil
+			return courier.ErrResponseUnexpected
```
rowanseymour, line 305: The send err types get converted to channel log errors... so the rule should be that you call `clog.Error` for non-terminal errors - i.e. something which is an error, but we'll carry on sending the message anyway

PR #698, opened 2024-02-21
Adjust channel logs for Twilio

norkans7:


handlers/nexmo/handler.go
```diff
@@ -227,15 +227,18 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
 		}
 
 		nexmoStatus, err := jsonparser.GetString(respBody, "messages", "[0]", "status")
+		errCode, _ := strconv.Atoi(nexmoStatus)
 		if err != nil || nexmoStatus != "0" {
-			return courier.ErrResponseUnexpected
+			// https://developer.vonage.com/messaging/sms/guides/troubleshooting-sms
+			clog.Error(courier.ErrorExternal("send:"+nexmoStatus, sendErrorCodes[errCode]))
+			return courier.ErrFailedWithReason(fmt.Sprint(errCode), sendErrorCodes[errCode])
```
rowanseymour, line 234: we don't want to these both because the error will be converted to a channel log error

rowanseymour (changes requested):

handlers/nexmo/handler_test.go
```diff
@@ -195,7 +195,9 @@ var defaultSendTestCases = []OutgoingTestCase{
 		ExpectedRequests: []ExpectedRequest{{
 			Form: url.Values{"text": {"Error status"}, "to": {"250788383383"}, "from": {"2020"}, "api_key": {"nexmo-api-key"}, "api_secret": {"nexmo-api-secret"}, "status-report-req": {"1"}, "type": {"text"}, "callback": {"https://localhost/c/nx/8eb23e93-5ecb-45ba-b726-3b064e0c56ab/status"}},
 		}},
-		ExpectedError: courier.ErrResponseUnexpected,
+
+		ExpectedError:     courier.ErrFailedWithReason("10", "Too Many Existing Binds"),
+		ExpectedLogErrors: []*courier.ChannelError{courier.ErrorExternal("send:10", "Too Many Existing Binds")},
```
rowanseymour, line 200: see above

handlers/twiml/handlers.go
```diff
@@ -291,7 +291,7 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
 				if errorCode == errorStopped {
 					return courier.ErrContactStopped
 				}
-				return courier.ErrResponseUnexpected
+				clog.Error(twilioError(errorCode))
```
rowanseymour, line 294:
```
return courier.ErrWithReason(...)
```

handlers/nexmo/handler.go
rowanseymour: the code here needs to be the way it was previously: `"send:"+nexmoStatus` to work with 

https://github.com/nyaruka/rapidpro/blob/832d6193a8bb9b5177bbe20d753f6ea2132d54fe/temba/channels/types/vonage/type.py#L117-L122

rowanseymour: the code here needs to be the way it was previously: "send:"+nexmoStatus to work with

https://github.com/nyaruka/rapidpro/blob/832d6193a8bb9b5177bbe20d753f6ea2132d54fe/temba/channels/types/vonage/type.py#L117-L122

rowanseymour (changes requested):

PR #703, opened 2024-02-22
Update D* handlers to use new send

norkans7:


handlers/dialog360/handler.go
```diff
@@ -549,7 +533,7 @@ func (h *handler) SendLegacy(ctx context.Context, msg courier.MsgOut, clog *cour
 
 					payload.Interactive = &interactive
 				} else {
-					return nil, fmt.Errorf("too many quick replies WAC supports only up to 10 quick replies")
+					return fmt.Errorf("too many quick replies WAC supports only up to 10 quick replies")
```
rowanseymour, line 536: should just truncate to 10 and log channel log error like we've done in other WhatsApp channels

rowanseymour (changes requested):

handlers/dialog360/handler.go
norkans7: Sure, I will add that here as well

PR #708, opened 2024-02-28
Rework WA handler to new send, remove channel legacy handler

norkans7:


handlers/test.go
```diff
@@ -408,29 +408,15 @@ func RunOutgoingTestCases(t *testing.T, channel courier.Channel, handler courier
...
-			legacyHandler, _ := handler.(courier.ChannelLegacyHandler)
-
 			var externalIDs []string
 			var status courier.StatusUpdate
-			var serr, err error
-
-			if legacyHandler != nil {
-				status, err = legacyHandler.SendLegacy(ctx, msg, clog)
+			var serr error
+			var resNewURN urns.URN
```
rowanseymour, line 414: since there's no more `if` `else` we don't need vars declared before they're assigned

rowanseymour (approved):

PR #711, opened 2024-03-07
Add channel log for error with line attachments

norkans7:


handlers/line/handler.go
```diff
@@ -317,6 +317,8 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
 
 		if err == nil {
 			jsonMsgs = append(jsonMsgs, string(jsonMsg))
+		} else {
+			clog.Error(courier.NewChannelError("", "", err.Error()))
```
norkans7, line 321: This line will help see why the attachments was not send

rowanseymour: if marshaling those things is creating an error.. .then just return that err (`return err`) and we'll fix it... but I'm not convinced this is where the problem is

rowanseymour (changes requested):

rowanseymour (approved):

PR #715, opened 2024-03-15
Fix wa button url

norkans7:


handlers/meta/whatsapp/templates.go
```diff
@@ -52,9 +52,9 @@ func GetTemplatePayload(templating MsgTemplating) *Template {
 		if strings.HasPrefix(k, "button.") {
 
 			for _, p := range v {
-				if strings.HasPrefix(p.Value, "http") {
+				if p.Type == "url" || strings.HasPrefix(p.Value, "http") {
 					component := &Component{Type: "button", Index: strings.TrimPrefix(k, "button."), SubType: "quick_reply"}
-					component.Params = append(component.Params, &Param{Type: "url", Text: p.Value})
+					component.Params = append(component.Params, &Param{Type: "text", Text: p.Value})
```
norkans7, line 57: https://developers.facebook.com/docs/whatsapp/cloud-api/reference/messages#button-parameter-object

type supported are payload and text

handlers/meta/whatsapp/templates.go
```diff
@@ -52,9 +52,9 @@ func GetTemplatePayload(templating MsgTemplating) *Template {
 		if strings.HasPrefix(k, "button.") {
 
 			for _, p := range v {
-				if strings.HasPrefix(p.Value, "http") {
+				if p.Type == "url" || strings.HasPrefix(p.Value, "http") {
```
norkans7, line 55: I update to have the url type for the param on rapidpro when syncing the templates
https://github.com/nyaruka/rapidpro/pull/5107

and the engine will pass that without changing it https://github.com/nyaruka/goflow/blob/b07f34a76065cea43070b0b0a2c509772a346ad8/flows/actions/send_msg.go#L168

rowanseymour: Why do we need `|| strings.HasPrefix(p.Value, "http")` ?

norkans7: That is what was there, so I did not want to remove that since the docs are not clear for every case,
I think that was added from trials and errors too

norkans7: But also I doubt that has ever worked since the we used payload and not text

norkans7: Actually I can remove that since the type url is an invalid value there

PR #716, opened 2024-03-15
Order components in template and fix max length for D3C

norkans7:


handlers/dialog360/handler.go
```diff
@@ -27,7 +27,7 @@ const (
 
 var (
 	// max for the body
-	maxMsgLength = 1000
+	maxMsgLength = 4096
```
norkans7, line 30: Not related but that is the correct max length

rowanseymour (approved):

PR #720, opened 2024-03-18
Fix button URLs for WA templates

norkans7:


handlers/meta/whatsapp/templates.go
```diff
@@ -92,6 +92,7 @@ func GetTemplatePayload(templating *MsgTemplating) *Template {
 
 			for _, p := range v {
 				if p.Type == "url" {
+					component.SubType = "url"
```
rowanseymour, line 95: Not convinced we've modeled this correctly since we now need to infer component type from param type... and these params should be just type `text`.. param type was just supposed to tell the editor what kinda of widget to display. Maybe `templating` should include component type?

norkans7: So we could add the component type to the components we save? in addition to the content and params keys?

norkans7: I tied adding the type to components https://github.com/nyaruka/rapidpro/pull/5111 however that will require we change the 

https://github.com/nyaruka/goflow/blob/545576482f9968644052b66f00544c972aa1cc33/flows/msg.go#L177

instead of 
```
Params_    map[string][]TemplateParam `json:"params,omitempty"`
```

We have a array of components

```
Components_ []TemplateComponen

type TemplateComponen struct {
   Name string // what was used as key in the previous map
   Type string
   Params []TemplateParam
}
```

And that could make the loop https://github.com/nyaruka/courier/blob/d9467e3e09748fc83d1276b26896eed4b2e5ad6d/handlers/meta/whatsapp/templates.go#L60 easier as well

PR #722, opened 2024-03-20
Use component name to find button index

norkans7:


norkans7:
Tests had kept the name field

https://github.com/nyaruka/courier/blob/f76d1c2d29ec33ce0b10b34f427269483f0869fa/handlers/meta/whatsapp/templates_test.go#L106 so no new tests needed for this

PR #723, opened 2024-03-27
Update payloads for tasks queued to mailroom

rowanseymour:


backends/rapidpro/backend_test.go
```diff
@@ -189,7 +189,7 @@ func (ts *BackendTestSuite) TestDeleteMsgByExternalID() {
 	err = ts.b.DeleteMsgByExternalID(ctx, knChannel, "ext2")
 	ts.Nil(err)
 
-	ts.assertQueuedContactTask(ContactID(100), "msg_deleted", map[string]any{"org_id": float64(1), "msg_id": float64(10002)})
```
rowanseymour, line 192: you don't need to repeat `org_id` or `contact_id` in the task payload... mailroom gets that from the queue

backends/rapidpro/task.go
```diff
@@ -32,15 +29,18 @@ func queueMsgHandling(rc redis.Conn, c *Contact, m *Msg) error {
 
 func queueChannelEvent(rc redis.Conn, c *Contact, e *ChannelEvent) error {
 	body := map[string]any{
-		"org_id":      e.OrgID_,
-		"contact_id":  e.ContactID_,
+		"event_id":    e.ID_,
+		"event_type":  e.EventType_,
```
rowanseymour, line 33: I want to switch these to all be a single `channel_event` task type, so step one is putting `event_type` in the payload

backends/rapidpro/task.go
```diff
@@ -32,15 +29,18 @@ func queueMsgHandling(rc redis.Conn, c *Contact, m *Msg) error {
 
 func queueChannelEvent(rc redis.Conn, c *Contact, e *ChannelEvent) error {
 	body := map[string]any{
-		"org_id":      e.OrgID_,
-		"contact_id":  e.ContactID_,
+		"event_id":    e.ID_,
```
rowanseymour, line 32: Seems like we should some day model whether channel events are handled.. i.e. have mailroom update their status like we do for messages.

norkans7 (approved):

backends/rapidpro/task.go
```diff
@@ -32,15 +29,18 @@ func queueMsgHandling(rc redis.Conn, c *Contact, m *Msg) error {
...
+		"event_type":  e.EventType_,
 		"urn_id":      e.ContactURNID_,
 		"channel_id":  e.ChannelID_,
 		"extra":       e.Extra(),
 		"new_contact": c.IsNew_,
 		"occurred_on": e.OccurredOn_,
 		"created_on":  e.CreatedOn_,
 	}
+	if e.OptInID_ != 0 {
+		body["optin_id"] = e.OptInID_
```
rowanseymour, line 42: best I can tell we're not doing optin triggers correctly in the engine because this isn't being passed - the triggers work but `@optin` is exposed in expressions

PR #740, opened 2024-05-08
Update gocommon

rowanseymour:


handlers/mblox/handler_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 var testChannels = []courier.Channel{
-	test.NewMockChannel("8eb23e93-5ecb-45ba-b726-3b064e0c56ab", "MB", "2020", "BR", map[string]any{"username": "zv-username", "password": "zv-password"}),
+	test.NewMockChannel("8eb23e93-5ecb-45ba-b726-3b064e0c56ab", "MB", "2020", "US", map[string]any{"username": "zv-username", "password": "zv-password"}),
```
rowanseymour, line 14: @norkans7 so seems `12067799294` is a valid local Brazilian number.. and if you pass `BR` then gocommon currently gives you back `+5512067799294` rather than `+12067799294`.. I think that's ok.. this is a weird test passing US numbers for BR channel. 

I guess if we know a channel is giving us international numbers we could manually add `+` before parsing... or don't pass the channel country because the channel in that case doesn't really have a country.

norkans7: Yes, the BR was from the test copied and pasted

PR #743, opened 2024-05-10
Add test to make sure URN are normalized for EX channel with non phon…

norkans7:
…e schemes

### DIFF
--- a/handlers/external/handler_test.go
+++ b/handlers/external/handler_test.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/nyaruka/courier/handlers"
 	"github.com/nyaruka/courier/test"
 	"github.com/nyaruka/gocommon/httpx"
+	"github.com/nyaruka/gocommon/urns"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


handlers/external/handler_test.go
```diff
@@ -267,11 +268,36 @@ var customTestCases = []IncomingTestCase{
...
+}
+
 func TestIncoming(t *testing.T) {
 	RunIncomingTestCases(t, testChannels, newHandler(), handleTestCases)
 	RunIncomingTestCases(t, testSOAPReceiveChannels, newHandler(), handleSOAPReceiveTestCases)
 	RunIncomingTestCases(t, gmChannels, newHandler(), gmTestCases)
 	RunIncomingTestCases(t, customChannels, newHandler(), customTestCases)
+
+	extChannel := test.NewMockChannel("8eb23e93-5ecb-45ba-b726-3b064e0c56ab", "EX", "2020", "GM", nil)
+	extChannel.SetScheme(urns.External.Prefix)
```
norkans7, line 299: I found this trick we already used in 
https://github.com/nyaruka/courier/blob/36bcdaa526dc44806481cc669380aadbda75b0e7/handlers/twiml/handlers_test.go#L1141
better instead of modifying `test.NewMockChannel` method to pass schemes

rowanseymour: Fine for this PR but really `NewMockChannel` should take schemes, we should set that in all tests, and we try to make our tests as realistic as possible

rowanseymour (approved):

PR #747, opened 2024-05-21
Use std library errors, part 1

norkans7:


handlers/hormuud/handler.go
```diff
@@ -79,7 +79,7 @@ type mtPayload struct {
 func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.SendResult, clog *courier.ChannelLog) error {
 	token, err := h.FetchToken(ctx, msg.Channel(), msg, clog)
 	if err != nil {
-		return errors.Wrapf(err, "unable to fetch token")
+		return fmt.Errorf("unable to fetch token")
```
rowanseymour, line 82: this isn't wrapping the error now

handlers/hormuud/handler_test.go
```diff
@@ -201,16 +198,8 @@ var tokenTestCases = []OutgoingTestCase{
...
-			{
-				Headers: map[string]string{
-					"Content-Type":  "application/json",
-					"Accept":        "application/json",
-					"Authorization": "Bearer ",
-				},
-				Body: `{"mobile":"250788383383","message":"Simple Message","senderid":"2020","mType":-1,"eType":-1,"UDH":""}`,
-			},
 		},
-		ExpectedError: courier.ErrResponseStatus,
```
rowanseymour, line 213: ?

handlers/hormuud/handler.go
norkans7: updated

handlers/hormuud/handler_test.go
norkans7: We are not reaching to the place we call the send mock since the token fails

handlers/hormuud/handler.go
```diff
@@ -79,7 +79,7 @@ type mtPayload struct {
 func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.SendResult, clog *courier.ChannelLog) error {
 	token, err := h.FetchToken(ctx, msg.Channel(), msg, clog)
 	if err != nil {
-		return errors.Wrapf(err, "unable to fetch token")
+		return courier.ErrFailedWithReason("", "unable to fetch token")
```
rowanseymour, line 82: the point of this method is to return an error with a code from the channel's side that can be used to help the user debug when viewing the channel log... what are you trying to do here?

rowanseymour (changes requested):

handlers/hormuud/handler.go
norkans7: return an error that can be matched in the tests

norkans7: Adjust to use STD errors

handlers/hormuud/handler.go
```diff
@@ -79,7 +79,7 @@ type mtPayload struct {
 func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.SendResult, clog *courier.ChannelLog) error {
 	token, err := h.FetchToken(ctx, msg.Channel(), msg, clog)
 	if err != nil {
-		return errors.Wrapf(err, "unable to fetch token")
+		return errors.New("unable to fetch token")
```
rowanseymour, line 82: why are you getting rid of the wrapping? I still don't undertstand why this isn't `fmt.Errorf("unable to fetch token: %w", err)` and why you need to change the tests....

norkans7:
```
actual  : *fmt.wrapError(&fmt.wrapError{msg:"unable to fetch token: error making token request: %!w(<nil>)", err:(*fmt.wrapError)(0x14000796100)})
```

norkans7: That is what we get when we run the test, how do we write the test to match that?

rowanseymour: the problem is we have an existing bug...

```go
if err != nil || resp.StatusCode/100 != 2 {
	return "", errors.Wrapf(err, "error making token request")
}
```

if `err == nil` but `resp.StatusCode` is non-200... that returns `"", nil`

PR #749, opened 2024-05-23
Fix hoormund bug so we don't ignore errors from fetching tokens

rowanseymour:


handlers/hormuud/handler_test.go
```diff
@@ -186,9 +186,6 @@ var tokenTestCases = []OutgoingTestCase{
 		MsgText: "Simple Message",
 		MsgURN:  "tel:+250788383383",
 		MockResponses: map[string][]*httpx.MockResponse{
-			"https://smsapi.hormuud.com/api/SendSMS": {
-				httpx.NewMockResponse(400, nil, []byte(`{"error": "invalid token"}`)),
-			},
```
rowanseymour, line 191: we shouldn't be making this call if fetching the token failed

norkans7: Yes

handlers/hormuud/handler.go
```diff
@@ -155,16 +151,18 @@ func (h *handler) FetchToken(ctx context.Context, channel courier.Channel, msg c
 	req.Header.Set("Accept", "application/json")
 
 	resp, respBody, err := h.RequestHTTP(req, clog)
-	if err != nil || resp.StatusCode/100 != 2 {
-		return "", errors.Wrapf(err, "error making token request")
+	if err != nil {
+		return "", courier.ErrConnectionFailed
+	} else if resp.StatusCode/100 != 2 {
+		return "", courier.ErrResponseStatus
```
rowanseymour, line 157: this check isn't quite right in so many handlers... maybe we should tweak `RequestHTTP` to return`ErrConnectionFailed` or  `ErrResponseStatus`... is there ever a case where we need to handle a non-200 response differently?

norkans7: We can check on that

norkans7 (approved):

handlers/hormuud/handler.go
rowanseymour: I guess the issue is that some handlers want to parse a 400 response to extract an error code

rowanseymour: Well either way I think you need to do an audit of `RequestHTTP` calls...

PR #752, opened 2024-06-06
Use split attachments for templates media attachment for consistency …

norkans7:
…with messages attachments

### DIFF
--- a/handlers/meta/whatsapp/templates.go
+++ b/handlers/meta/whatsapp/templates.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/nyaruka/courier"
+	"github.com/nyaruka/courier/handlers"
 	"golang.org/x/exp/maps"
 )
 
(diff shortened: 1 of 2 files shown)


handlers/meta/whatsapp/templates.go
```diff
@@ -30,18 +31,26 @@ func GetTemplatePayload(templating *courier.Templating) *Template {
...
-				} else if p.Type == "document" {
-					component.Params = append(component.Params, &Param{Type: p.Type, Document: &struct {
-						Link string "json:\"link,omitempty\""
-					}{Link: p.Value}})
+				if p.Type != "text" {
+					attType, attURL := handlers.SplitAttachment(p.Value)
+					attType = strings.Split(attType, "/")[0]
+					if attType == "application" {
+						attType = "document"
+					}
```
rowanseymour, line 39: I think this code is simpler to read if we let `attType` stay as "application".. because that _is_ is the attachment type in our world, and use `document` in the WhatsApp payload

rowanseymour (approved):

PR #753, opened 2024-06-06
Adjust attachment variables for TWA templates

norkans7:
matches the attachments for TWA channel as for #752

### DIFF
--- a/handlers/twiml/handlers.go
+++ b/handlers/twiml/handlers.go
@@ -253,7 +253,13 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
 
 		for _, comp := range msg.Templating().Components {
 			for varKey, varIndex := range comp.Variables {
-				contentVariables[varKey] = msg.Templating().Variables[varIndex].Value
+				value := msg.Templating().Variables[varIndex].Value
+
+				if msg.Templating().Variables[varIndex].Type != "text" {
+					_, value = handlers.SplitAttachment(value)
+				}
+
+				contentVariables[varKey] = value
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


rowanseymour (approved):

handlers/twiml/handlers.go
```diff
@@ -253,7 +253,16 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
 
 		for _, comp := range msg.Templating().Components {
 			for varKey, varIndex := range comp.Variables {
-				contentVariables[varKey] = msg.Templating().Variables[varIndex].Value
+				value := msg.Templating().Variables[varIndex].Value
+				attType, value := handlers.SplitAttachment(value)
+				attType = strings.Split(attType, "/")[0]
+
+				// not attachment ignore the split, revert to original entire value
```
norkans7, line 260: @rowanseymour hold on I found a better way here

handlers/twiml/handlers.go
```diff
@@ -253,7 +253,16 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
...
+				if attType != "image" && attType != "video" && attType != "application" {
+					value = msg.Templating().Variables[varIndex].Value
+				}
```
rowanseymour, line 263: actually this feels a bit weird and is liable to do the wrong thing if a text variable looks like an attachment... can't we check variable type like in the WhatsApp handler ?

rowanseymour (changes requested):

handlers/twiml/handlers.go
norkans7: checking the type `msg.Templating().Variables[varIndex].Type` so we can split as attachment

handlers/twiml/handlers.go
norkans7: Yes, I had found that just after pushing

rowanseymour (approved):

PR #764, opened 2024-07-25
Update AWS/S3 config

rowanseymour:


backends/rapidpro/backend.go
```diff
@@ -174,22 +174,16 @@ func (b *backend) Start() error {
 		s3config := &storage.S3Options{
 			AWSAccessKeyID:     b.config.AWSAccessKeyID,
 			AWSSecretAccessKey: b.config.AWSSecretAccessKey,
+			Region:             b.config.AWSRegion,
 			Endpoint:           b.config.S3Endpoint,
-			Region:             b.config.S3Region,
-			DisableSSL:         b.config.S3DisableSSL,
```
rowanseymour, line 179: Best I can tell this isn't needed if you're providing an endpoint as a full URL with http or https

backends/rapidpro/backend.go
```diff
@@ -174,22 +174,16 @@ func (b *backend) Start() error {
...
-			ForcePathStyle:     b.config.S3ForcePathStyle,
 			MaxRetries:         3,
 		}
-		if b.config.AWSAccessKeyID != "" && !b.config.AWSUseCredChain {
-			s3config.AWSAccessKeyID = b.config.AWSAccessKeyID
-			s3config.AWSSecretAccessKey = b.config.AWSSecretAccessKey
```
rowanseymour, line 185: These are being set above anyway. The logic should be that if `AWSUseCredChain` is set, you get S3 storage if tho `AWSAccessKeyID` and `AWSSecretAccessKey` are blank.

backends/rapidpro/backend.go
```diff
@@ -655,7 +649,7 @@ func (b *backend) SaveAttachment(ctx context.Context, ch courier.Channel, conten
 
 	orgID := ch.(*Channel).OrgID()
 
-	path := filepath.Join(b.config.S3AttachmentsPrefix, strconv.FormatInt(int64(orgID), 10), filename[:4], filename[4:8], filename)
+	path := filepath.Join("attachments", strconv.FormatInt(int64(orgID), 10), filename[:4], filename[4:8], filename)
```
norkans7, line 652: Why not keep that as `media`?
Are the files actually moved on that path now?

rowanseymour: This is just what is set on the default config (used by tests) that we always overwrite on prod to attachments. RP, mailroom and courier should all use the same `attachments` prefix.

norkans7 (approved):

PR #777, opened 2024-09-04
Added quick replies data for fcm channel 

eamin-dev:
This pull request introduces functionality to handle quick replies data as string for the FCM (Firebase Cloud Messaging) channel. The changes include:

- **Conditional Check**: Added a check to determine if quick replies are present in the message.

- **JSON Serialization**: Implemented JSON serialization for the quick replies data using jsonx.MustMarshal.

- **Payload Update**: Updated the FCM payload to include the serialized quick replies data.

These updates ensure that quick replies are properly included in the FCM message payload, improving message interactivity and user engagement.

### DIFF
--- a/handlers/firebase/handler.go
+++ b/handlers/firebase/handler.go
@@ -138,6 +138,7 @@ type mtPayload struct {
 			Message       string `json:"message"`
 			MessageID     string `json:"message_id"`
 			SessionStatus string `json:"session_status"`
+			QuickReplies  string `json:"quick_replies,omitempty"`
 		} `json:"data"`
 		Notification *mtNotification `json:"notification,omitempty"`
 		Token        string          `json:"token"`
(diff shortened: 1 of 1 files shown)


handlers/firebase/handler.go
```diff
@@ -286,6 +287,10 @@ func (h *handler) sendWithCredsJSON(msg courier.MsgOut, res *courier.SendResult,
 
 		payload.Message.Token = msg.URNAuth()
 		payload.Message.Android.Priority = "high"
+		
+		if quickReplies := msg.QuickReplies(); quickReplies != nil {
+			payload.Message.Data.QuickReplies = string(jsonx.MustMarshal(quickReplies))
```
norkans7, line 292: I understand you chose to expect a string with the quick replies joined, however this PR is not quite right as the quick replies here would be sent for each part of the message.
We want that on the last part only

norkans7:
You should have added some tests and signed the CLA

however I had already added a draft PR assuming you will chose to expect a string with the quick replies joined
So I will merge that https://github.com/nyaruka/courier/pull/776/files

norkans7:
replaced by https://github.com/nyaruka/courier/pull/776/files

PR #786, opened 2024-09-13
Replace variables with quotes

norkans7:
This will allow support both 

The current expected config format for variables
`{'body': '{"from":{{from_no_plus}},"to":{{to_no_plus}},"text":{{text}} }', 'method': 'POST', 'encoding': 'D', 'send_url': 'XXXXXXXXX', 'max_length': 160, 'content_type': 'json', 'mt_response_check': '', 'send_authorization': ''}`

And not generate the wrong JSON for the following with quotes around variables as below
`{'body': '{"from":"{{from_no_plus}}","to":"{{to_no_plus}}","text":"{{text}}" }', 'method': 'POST', 'encoding': 'D', 'send_url': 'XXXXXXXXX', 'max_length': 160, 'content_type': 'json', 'mt_response_check': '', 'send_authorization': ''}`

Or even 
`{'body': '{"from":'{{from_no_plus}}',"to":'{{to_no_plus}}',"text":"{{text}}" }', 'method': 'POST', 'encoding': 'D', 'send_url': 'XXXXXXXXX', 'max_length': 160, 'content_type': 'json', 'mt_response_check': '', 'send_authorization': ''}`

### DIFF
--- a/handlers/external/handler.go
+++ b/handlers/external/handler.go
@@ -429,6 +429,8 @@ func encodeVariables(variables map[string]string, contentType string) map[string
 
 func replaceVariables(text string, variables map[string]string) string {
 	for k, v := range variables {
+		text = strings.Replace(text, fmt.Sprintf("\"{{%s}}\"", k), v, -1)
+		text = strings.Replace(text, fmt.Sprintf("'{{%s}}'", k), v, -1)
 		text = strings.Replace(text, fmt.Sprintf("{{%s}}", k), v, -1)
 	}
 	return text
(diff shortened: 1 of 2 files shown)


rowanseymour:
@norkans7 I'm not following why this is needed.. why would someone put `"to":'{{to_no_plus}}'` in JSON?

norkans7:
I noticed a user that added an EX channel type with those request body and we generated an invalid JSON body.
I think we could try hard to make those work by replacing the part and including the quotes

rowanseymour:
Can't we just validate the JSON on the RP side? It seems weird to let them save nonsense like single quote strings.

norkans7:
The request body for EX channel type is not always JSON, and what we expect current is not valid JSON either,
I think being flexible is helpful when we replace the variables

norkans7:
Replaced by https://github.com/nyaruka/rapidpro/pull/5487

PR #805, opened 2024-11-25
Add send error for channel and contact pair limit hit that allow retr…

norkans7:
…ying the message

### DIFF
--- a/handlers/meta/handlers.go
+++ b/handlers/meta/handlers.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


handlers/meta/handlers.go
```diff
@@ -1090,6 +1091,10 @@ func (h *handler) requestWAC(payload whatsapp.SendRequest, accessToken string, r
 		return courier.ErrResponseUnparseable
 	}
 
+	if slices.Index([]int{4, 80007, 130429, 131048, 131056, 133016}, respPayload.Error.Code) > 0 {
```
norkans7, line 1094: Matching the error codes https://developers.facebook.com/docs/whatsapp/cloud-api/support/error-codes#throttling-errors to all be marked as throttle error and we retry up to 3 times

rowanseymour: this is odd way to check if x in y. Save the list of error codes as a global, then use `slices.Contains`

rowanseymour (changes requested):

rowanseymour (approved):

PR #807, opened 2024-12-04
Add support for document send with link when attachment type is document

norkans7:


handlers/dialog360/handler_test.go
```diff
@@ -386,6 +386,24 @@ var SendTestCasesD3C = []OutgoingTestCase{
 		},
 		ExpectedExtIDs: []string{"157b5e14568e8"},
 	},
+	{
```
norkans7, line 389: Just tests to make sure the issue is not happening fro this channel type 

https://github.com/nyaruka/courier/blob/c211b9c76b179b6352256bc0f95c977e2f82a2f1/handlers/dialog360/handler.go#L408

rowanseymour (approved):

PR #808, opened 2024-12-06
Fix WA template document  name

norkans7:


handlers/meta/whatsapp/api.go
```diff
@@ -149,7 +149,8 @@ type Param struct {
 		Link string `json:"link,omitempty"`
 	} `json:"video,omitempty"`
 	Document *struct {
-		Link string `json:"link,omitempty"`
+		Link     string `json:"link,omitempty"`
+		Filename string `json:"filename,omitempty"`
```
norkans7, line 153: https://developers.facebook.com/docs/whatsapp/cloud-api/reference/messages#media-object

Filename should be used for document type only

PR #809, opened 2024-12-12
Update Facebook API version to v18.0

norkans7:


attachments.go
```diff
@@ -155,6 +156,10 @@ func getAttachmentType(t *httpx.Trace) (string, string) {
 		extensions, err := mime.ExtensionsByType(typ)
 		if len(extensions) > 0 && err == nil {
 			ext = extensions[0][1:]
+			if slices.Contains([]string{"jpe", "jfif"}, ext) {
```
norkans7, line 159: This was returning different extensions but they are the same format so I used this workaround until I find why the mimetypes dependency we use is not consistent there

PR #810, opened 2024-12-13
Add cloudwatch config variables

norkans7:


norkans7:
included in https://github.com/nyaruka/courier/pull/813

PR #818, opened 2024-12-17
Refactor metrics so that everything is sent from Heartbeat in the backend

rowanseymour:


backend.go
```diff
@@ -106,9 +106,6 @@ type Backend interface {
 
 	// RedisPool returns the redisPool for this backend
 	RedisPool() *redis.Pool
-
-	// CloudWatch return the CloudWatch service for this backend
-	CloudWatch() *cwatch.Service
```
rowanseymour, line 111: Cloudwatch is now completely internal to the RP backend.. which seems right

backends/rapidpro/backend.go
```diff
@@ -770,38 +775,32 @@ func (b *backend) Heartbeat() error {
...
+	b.dbWaitDuration = dbStats.WaitDuration
+	b.redisWaitDuration = redisStats.WaitDuration
 
 	hostDim := cwatch.Dimension("Host", b.config.InstanceID)
-
-	b.CloudWatch().Queue(
+	metrics = append(metrics,
 		cwatch.Datum("DBConnectionsInUse", float64(dbStats.InUse), cwtypes.StandardUnitCount, hostDim),
-		cwatch.Datum("DBConnectionWaitDuration", float64(dbWaitDurationInPeriod/time.Millisecond), cwtypes.StandardUnitMilliseconds, hostDim),
+		cwatch.Datum("DBConnectionWaitDuration", float64(dbWaitDurationInPeriod/time.Second), cwtypes.StandardUnitSeconds, hostDim),
```
rowanseymour, line 789: we're using seconds everywhere

backends/rapidpro/backend.go
```diff
@@ -131,6 +129,8 @@ func newBackend(cfg *courier.Config) courier.Backend {
 		receivedExternalIDs: redisx.NewIntervalHash("seen-external-ids", time.Hour*24, 2), // 24 - 48 hours
 		sentIDs:             redisx.NewIntervalSet("sent-ids", time.Hour, 2),              // 1 - 2 hours
 		sentExternalIDs:     redisx.NewIntervalHash("sent-external-ids", time.Hour, 2),    // 1 - 2 hours
+
+		stats: NewStatsCollector(),
```
norkans7, line 133: Do we need to call Heartbeat when the backend stop or in backend cleanup so we are sure all the stats at that time are sent? it seems the stats in the last minute not yet send will be lost

norkans7 (approved):

PR #826, opened 2025-01-09
Do not log EX MT response check error

norkans7:


handlers/external/handler.go
```diff
@@ -366,7 +366,7 @@ func (h *handler) Send(ctx context.Context, msg courier.MsgOut, res *courier.Sen
 		}
 
 		if responseCheck != "" && !strings.Contains(string(respBody), responseCheck) {
-			return courier.ErrResponseUnexpected
+			return courier.ErrFailedWithReason("mt-response-check", fmt.Sprintf("response not containing '%s'", responseCheck))
```
rowanseymour, line 369: Hmm I think we want basically the same as `ErrResponseUnexpected` but not logged.. so maybe just a new error constant:

```go
var ErrResponseUnexpectedUnlogged error = &SendError{
	msg:       "response not expected values",
	retryable: false,
	loggable:  false,
	clogCode:  "response_unexpected",
	clogMsg:   "Response doesn't match expected values.",
}
```

sender.go
```diff
@@ -110,6 +110,15 @@ var ErrResponseUnexpected error = &SendError{
 	clogMsg:   "Response doesn't match expected values.",
 }
 
+// ErrResponseUnexpectedUnlogged should be returned when channel response doesn't match what we expect and unlogged
```
rowanseymour, line 113: // ErrResponseUnexpectedUnlogged is same as ErrResponseUnexpected without logging so better for channels where response check is user configured.

rowanseymour (changes requested):

PR #834, opened 2025-02-12
Add delete request endpoint that creates a channel event for FBA chan…

norkans7:
…nels

### DIFF
--- a/handlers/meta/handlers.go
+++ b/handlers/meta/handlers.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/buger/jsonparser"
+	"github.com/getsentry/sentry-go"
 	"github.com/nyaruka/courier"
 	"github.com/nyaruka/courier/handlers"
 	"github.com/nyaruka/courier/handlers/meta/messenger"
@@ -85,6 +86,7 @@ func (h *handler) Initialize(s courier.Server) error {
 	h.SetServer(s)
 	s.AddHandlerRoute(h, http.MethodGet, "receive", courier.ChannelLogTypeWebhookVerify, h.receiveVerify)
 	s.AddHandlerRoute(h, http.MethodPost, "receive", courier.ChannelLogTypeMultiReceive, handlers.JSONPayload(h, h.receiveEvents))
+	s.AddHandlerRoute(h, http.MethodPost, "delete", courier.ChannelLogTypeEventReceive, handlers.JSONPayload(h, h.deleteEvents))
 	return nil
 }
 
@@ -130,7 +132,7 @@ func (h *handler) WriteRequestError(ctx context.Context, w http.ResponseWriter, 
 
 // GetChannel returns the channel
 func (h *handler) GetChannel(ctx context.Context, r *http.Request) (courier.Channel, error) {
-	if r.Method == http.MethodGet {
+	if r.Method == http.MethodGet || r.URL.Path == "/c/fba/delete" {
 		return nil, nil
 	}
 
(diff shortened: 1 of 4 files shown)


rowanseymour:
@norkans7 can you put the changes to write event UUIDs in a separate PR?

norkans7:
@rowanseymour I moved that to https://github.com/nyaruka/courier/pull/835 and rebased this on that branch

norkans7:
QA as

norkans7:
contains #846

handlers/meta/handlers.go
```diff
@@ -215,6 +217,58 @@ func (h *handler) resolveMediaURL(mediaID string, token string, clog *courier.Ch
...
+		return nil, handlers.WriteAndLogRequestIgnored(ctx, h, channel, w, r, "ignoring request, no existing contact matched")
+	}
+
+	date := parseTimestamp(payload.IssuedAt)
+
+	events := make([]courier.Event, 0, 2)
+	data := make([]any, 0, 2)
+
+	payloadJson, _ := json.Marshal(payload)
+	sentry.CaptureMessage(fmt.Sprintf("Data Deletion Request: %s", payloadJson))
```
rowanseymour, line 255: I think this should be on the mailroom side

handlers/meta/handlers.go
```diff
@@ -215,6 +217,58 @@ func (h *handler) resolveMediaURL(mediaID string, token string, clog *courier.Ch
...
+
+	event := h.Backend().NewChannelEvent(channel, courier.EventDeleteContact, urn, clog).WithOccurredOn(date).WithExtra(map[string]string{"userID": payload.UserID})
+
+	err = h.Backend().WriteChannelEvent(ctx, event, clog)
+	if err != nil {
+		return nil, err
+	}
+
+	confirmationURL := fmt.Sprintf("https://%s/channels/events/read/%s/", h.Server().Config().Domain, event.UUID())
```
rowanseymour, line 264: @norkans7 @ericnewcomer need to make a decision on this URL.. is it maybe odd to have event UUID in the URL and be the confirmation code? Should we use contact UUID instead? Is there a vulnerability there allowing enumeration of contact UUIDs? My inclination is a really generic URL like `https://<domain>/public/forgetme/` with a UUID as the confirmation code that they enter.

norkans7: I find that once we submit the form, we need a page to redirect to so i think we need to keep the event read view for that.
And that approach gives us flexibility to change that behavior with changing the link we send for the confirmation

norkans7: I used the same view with a get param
https://github.com/nyaruka/rapidpro/pull/5928/files

handlers/meta/handlers.go
```diff
@@ -130,7 +131,7 @@ func (h *handler) WriteRequestError(ctx context.Context, w http.ResponseWriter,
 
 // GetChannel returns the channel
 func (h *handler) GetChannel(ctx context.Context, r *http.Request) (courier.Channel, error) {
-	if r.Method == http.MethodGet {
+	if r.Method == http.MethodGet || r.URL.Path == "/c/fba/delete" {
```
rowanseymour, line 134: @norkans7 I'm not getting why we need this..

norkans7: We use the same URL for all channels and this lookup the channel for the message however for the delete request we do not get info about the channel at all to be able to look it up so will ignore and the URN we get will be unique since they page scoped

rowanseymour: but how can we look up a URN without a channel? Don't we need channel to get org_id ?

norkans7: Right, this will not work without a valid channel, the request I expect we get on that also cannot allow us to lookup the channel

I guess we remove the check for the contacts exists in the DB and rely on the URN that we know if page scoped

rowanseymour: We can't create a channel event without a channel... and we don't gain much from making it mailroom's problem to determine if a URN exists - we don't have an index on URN without org id. Argh.

PR #841, opened 2025-02-26
Messagebird Handler Updates

tybritten:
Updated to support Messagebird's default webhook receiving handlers, with both shortcodes and regular numbers.

### DIFF
--- a/handlers/messagebird/handler.go
+++ b/handlers/messagebird/handler.go
@@ -11,7 +11,7 @@ import (
 	"encoding/hex"
 	"log/slog"
 	"strconv"
-
+	"errors"
 	"fmt"
 
 	"net/http"
@@ -62,14 +62,16 @@ var statusMapping = map[string]courier.MsgStatus{
 	"expired":         courier.MsgStatusFailed,
 }
 
-type ReceivedMessage struct {
-	ID              string   `json:"id"`
-	Recipient       string   `json:"recipient"`
-	Originator      string   `json:"originator"`
-	Body            string   `json:"body"`
-	CreatedDatetime string   `json:"createdDatetime"`
-	MediaURLs       []string `json:"mediaUrls"`
-	MMS             bool     `json:"mms"`
+type formMessage struct {
+	ID              string   `name:"id"`
+	MID             string   `name:"mid"`  //shortcode only
+	Shortcode       string   `name:"shortcode"`  //shortcode only	
+	Recipient       string   `name:"recipient" validate:"required"` 
+	Originator      string   `name:"originator"`
+	Body            string   `name:"body"`
+	MediaURLs       []string `name:"mediaUrls"`
+	MessageBody     string   `name:"message"` //shortcode only
+	CreatedDatetime string   `name:"receive_datetime" validate:"required"`
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


handlers/messagebird/handler.go
```diff
@@ -62,14 +62,16 @@ var statusMapping = map[string]courier.MsgStatus{
...
-	ID              string   `json:"id"`
-	Recipient       string   `json:"recipient"`
-	Originator      string   `json:"originator"`
-	Body            string   `json:"body"`
-	CreatedDatetime string   `json:"createdDatetime"`
-	MediaURLs       []string `json:"mediaUrls"`
-	MMS             bool     `json:"mms"`
+type formMessage struct {
+	ID              string   `name:"id"`
+	MID             string   `name:"mid"`  //shortcode only
```
rowanseymour, line 67: doesn't look like you're using gofmt.. we're all using vscode set to use default go formatter

tybritten: ah I used to, it must be turned off. I'll fix.

handlers/messagebird/handler.go
```diff
@@ -62,14 +62,16 @@ var statusMapping = map[string]courier.MsgStatus{
 	"expired":         courier.MsgStatusFailed,
 }
 
-type ReceivedMessage struct {
```
norkans7, line 65: Why not keep expecting the JSON format for this handler?

handlers/messagebird/handler.go
```diff
@@ -62,14 +62,16 @@ var statusMapping = map[string]courier.MsgStatus{
...
+type formMessage struct {
+	ID              string   `name:"id"`
+	MID             string   `name:"mid"`       //shortcode only
+	Shortcode       string   `name:"shortcode"` //shortcode only
+	Recipient       string   `name:"recipient" validate:"required"`
+	Originator      string   `name:"originator"`
+	Body            string   `name:"body"`
+	MediaURLs       []string `name:"mediaUrls"`
+	MessageBody     string   `name:"message"` //shortcode only
+	CreatedDatetime string   `name:"receive_datetime" validate:"required"`
```
norkans7, line 74: We need a different field for the date time `createdDatetime ` for long numbers  and `receive_datetime` for short codets , we can remove the required tag 

https://developers.messagebird.com/api/sms-messaging/#receive-and-handle-inbound-sms

handlers/messagebird/handler.go
tybritten: the default webhook on MB is urlencoded, we were using a internal flow with them that turned it to json (among other things) before we found out they charge extra for these.

handlers/messagebird/handler.go
tybritten: ah yes, I need to fix that- this is the annoying part where they name the fields differently between shortcodes and long numbers

handlers/messagebird/handler.go
norkans7: Changing that will require you update each existing callback to use the default urlencoded again, I imagine those are now set to use JSON when submitting

tybritten: surprisingly sending with MB is either json or urlencoded, so that doesn't need to change. Since we only use them for shortcodes, we have a patch of our current courier with some of these changes just so we could move back to their default webhook to cut our SMS costs

norkans7: I think we can keep expecting JSON, and make that clear on the config page https://github.com/nyaruka/rapidpro/blob/main/temba/channels/types/messagebird/type.py#L35 that the callback should be configured with JSON

rowanseymour: Let's take a look and see if we have customers using messagebird that way... it would be nice to support whatever the default setup is for this channel.

tybritten: its basically like the difference between using the twilio webhook, or using a twilio function to convert the message to json. there's a cost associated that they don't make clear. the actual webhook from the number is always form encoded, you have to manually set up a flow (function) to convert. We did that before we knew there was a cost per run of that flow.

rowanseymour: Ok we have no channels of this type in prod probably because the type has always been beta gated. Since the default way to connect this channel sends urlencoded params, that's what we should support. And when we've merged this and tested it.. let's un- beta gate it.

PR #865, opened 2025-06-12
Pass context to the task queueing

norkans7:


rowanseymour:
I don't think this fixes the queueing issue but we should merge it because we should be using context with valkey calls

PR #873, opened 2025-07-08
Queue to new mailroom realtime queue

rowanseymour:


backends/rapidpro/backend_test.go
```diff
@@ -1544,17 +1544,17 @@ func (ts *BackendTestSuite) assertNoQueuedContactTask(contactID ContactID) {
 	rc := ts.b.rp.Get()
 	defer rc.Close()
 
-	assertvk.ZCard(ts.T(), rc, "tasks:handler:1", 0)
-	assertvk.ZCard(ts.T(), rc, "tasks:handler:active", 0)
+	assertvk.ZCard(ts.T(), rc, "{tasks-realtime:1}", 0)
+	assertvk.ZCard(ts.T(), rc, "{tasks-realtime}:queued", 0)
```
norkans7, line 1548: why is `}` not after `:queued` ?

rowanseymour: Read up on hashtags at https://valkey.io/topics/cluster-spec/ if you haven't already. The part of key between the `{` and `}` is what determines the hash slot. For a multi-key structure like `queues.Fair` we don't need all keys to be in the same hash slot - just the keys that are accessed together. So `{tasks-realtime}:queued` and `{tasks-realtime}:active` need matching hash tag portion, but `{tasks-realtime:1}` (the task queue for org 1) can be in a separate hash slot... each task queue for each different owner can be in a different slot thus spreading the load out better.

rowanseymour: Actually `{tasks-realtime:1}` is wrong.. it should be `{tasks-realtime:1}/0`..

PR #874, opened 2025-07-29
Add new DynamoDB writer and spool for channel logs

rowanseymour:


ericnewcomer:
I really appreciate the summary you get from a copilot review, even if the review doesn't suggest anything actionable.

PR #1032, opened 2026-06-02
Replace httpx.DoTrace with WithTracing/WithBodyLimit transports

rowanseymour:
Migrates outbound HTTP tracing off `httpx.DoTrace` onto the composable transport middleware in gocommon v1.81.x, completing the move started with `WithAccessControl`/`WithMocks`. After this, courier makes no `httpx.DoTrace` calls. Targets gocommon **v1.81.3** (which dropped `WithTracing`'s `maxBodyBytes` param and renamed the new constructors: `WithTracing→WithTraces`, `WithBodyLimit→WithReadLimit`, `WithMocking→WithMocks`).

## `utils.TraceHTTP`

New helper that layers tracing onto a client's transport *for a single call* — so concurrent senders never race on shared trace state — while keeping any existing middleware (e.g. access control) in effect:

```go
httpx.WithTraces(httpx.WithReadLimit(client.Transport, maxBodyBytes))
```

- `WithReadLimit` bounds the bytes read from each response body (rejecting larger bodies with `ErrResponseSize` — the protection needed for untrusted endpoints); `WithTraces` captures the resulting bounded body into the trace. `maxBodyBytes <= 0` reads/captures the whole body.
- `WithTraces` defers a body-read error onto the response body, so `TraceHTTP` drains the response (when a limit is set) to surface `ErrResponseSize` as the returned error, exactly as `DoTrace` reported it.
- On a redirect, only the **final hop's** trace is returned (intermediate redirect hops are dropped), so a request yields one channel-log entry as `DoTrace` did — not one per hop. This is done by position (the hop `client.Do` returned), not by filtering 3xx, so a `304`/other unfollowed 3xx, or a redirect loop, is still logged.

It lives in `utils` (it only depends on `net/http`/`httpx`, no courier-package types).

## Call sites

- **`handlers/base.go`** — the `requestHTTP` chokepoint every handler uses now traces via `utils.TraceHTTP`.
- **`attachments.go`** — migrated too: `WithReadLimit` supplies the 100 MB read bound that previously kept it on `DoTrace`, so it's no longer the lone exception.
- **`test/handler.go`** — the mock handler migrated the same way.
- **Tests** (`channel_log_test`, `server_test`, `backend_test`) build their traces via `utils.TraceHTTP`; `clog_test` uses an inline `httpx.WithTraces` closure to avoid the `clogs` package test importing its parent. Mocking moved from the global `SetRequestor` to transport-scoped `httpx.WithMocks`.

(The remaining `httpx.Do` in the VK handler is the non-tracing raw media download — out of scope here.)

## Testing

- `utils.TestTraceHTTP` covers within-limit capture, oversize → `ErrResponseSize`, the no-limit case, and the redirect-yields-one-trace case.
- `go vet ./...`
- `go test -p=1 ./...` — all 63 packages pass


rowanseymour:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1053, opened 2026-07-07
Support WhatsApp BSUIDs as whatsapp URNs alongside phone numbers

norkans7:
A whatsapp URN may now hold either a phone number (all digits) or a business-scoped user ID (the CC.xxx form), so:

- sending routes the URN to the correct WhatsApp API field by content - a business-scoped user ID goes in the recipient field and a phone number in the to field (previously keyed on the separate bsuid scheme)
- receiving stores an incoming user_id/BSUID as a whatsapp URN (was the bsuid scheme), makes it the message's primary URN, and attaches the phone number as a secondary URN
- contact resolution looks up an existing contact by the whatsapp BSUID first and falls back to the phone number's all-digit whatsapp URN, reusing a contact that predates the BSUID instead of duplicating it

Applied across the meta, dialog360, turn and twiml handlers plus the rapidpro backend. Depends on the gocommon change accepting business-scoped user IDs as WhatsApp URNs; a dev-only replace directive points at it until that is released.

### DIFF
--- a/backends/rapidpro/backend_test.go
+++ b/backends/rapidpro/backend_test.go
@@ -985,6 +985,25 @@ func (ts *BackendTestSuite) TestContactForMsg() {
 	ts.NoError(err)
 	ts.False(matchedByBSUID.IsNew_)
 	ts.Equal(c1.ID_, matchedByBSUID.ID_)
+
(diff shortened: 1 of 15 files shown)


norkans7:
Replaced by https://github.com/nyaruka/courier/pull/1061

PR #1060, opened 2026-07-14
Replace unmaintained patrickmn/go-cache with gocommon's cache.Local

rowanseymour:
The only usage was `failedMediaCache` in the Turn.io WhatsApp handler — a negative cache of media URLs that recently failed to upload. `patrickmn/go-cache` is effectively unmaintained (last release 2017) and pre-generics, while gocommon's `cache.Local` is already used elsewhere in courier.

Semantics are unchanged: 15-minute TTL per entry, expired entries are never returned, and a failed upload suppresses retries for 15 minutes. The background sweeper (`Start()`) replaces go-cache's janitor.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,6 @@ require (
 	github.com/nyaruka/goflow v0.280.0
 	github.com/nyaruka/null/v3 v3.0.0
 	github.com/nyaruka/vkutil v0.21.0
-	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/samber/slog-multi v1.8.0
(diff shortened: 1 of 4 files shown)


rowanseymour:
Confirmed: `cache.Local.Get` delegates to ttlcache v3's `Get`, whose internal `get` checks `item.isExpiredUnsafe()` synchronously and returns nil for expired items regardless of whether the background sweeper has run. So expired entries are never returned and the retry-suppression window is exactly the 15-minute TTL, matching go-cache's behavior.

PR #1061, opened 2026-07-15
Support WhatsApp BSUIDs as whatsapp URNs alongside phone numbers

norkans7:
A WhatsApp identity can be either a phone number or a business-scoped user ID (BSUID). Store both under the single whatsapp URN scheme rather than a separate scheme.

- Sending: route a BSUID to the recipient field and a phone number to the to field.
- Receiving: when a message also carries a BSUID, make that whatsapp URN the primary URN and attach the phone number as a secondary URN.
- Contact resolution: look an existing contact up by the BSUID first, falling back to the phone number's whatsapp URN; a contact matched by phone has the BSUID added to it, without stealing a URN that already belongs to another contact.

### DIFF
--- a/backends/rapidpro/backend_test.go
+++ b/backends/rapidpro/backend_test.go
@@ -994,10 +994,12 @@ func (ts *BackendTestSuite) TestContactForMsg() {
 	ts.NoError(err)
 	ts.NotEqual(bsuidOwner.ID_, phoneOwner.ID_)
 
-	resolved, err := contactForMsg(ctx, ts.b, newBSUIDMsg("whatsapp:US.7777", "whatsapp:12065557777"), clog)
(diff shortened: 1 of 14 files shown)


rowanseymour:
The sending side of this (routing a BSUID to the recipient field and a phone number to the to field, across the WAC/D3C/TRN/TWA handlers) has now landed via #1083, which also covers BSUIDs held as whatsapp URNs.

The receive side — making the BSUID the primary URN and the phone/BSUID contact matching — we're going to take a different route on: BSUIDs will be saved as secondary whatsapp URNs without displacing the phone as the primary, so incoming contact resolution by phone stays untouched for now. That work will come in a follow-up PR, so closing this one.

PR #1063, opened 2026-07-15
Update gocommon to v1.89.5

rowanseymour:
Updates `github.com/nyaruka/gocommon` from v1.89.4 to v1.89.5.

That release bumps `github.com/nyaruka/phonenumbers/v2` from v2.0.2 to v2.0.4, which fixes two denial-of-service issues:

- a slice-bounds panic in `Parse` on malformed RFC3966 input
- an unbounded allocation in `GetNationalSignificantNumber` / `Format` / `IsValidNumber` from an untrusted `PhoneNumber`

Both are reachable here transitively via gocommon's `urns` and `i18n` packages, which parse phone numbers from inbound channel requests.

Version bump only — no import path or API changes. Builds clean and the test suite passes.


rowanseymour:
Superseded by #1066, which updated gocommon to v1.89.7 on main. That release includes the phonenumbers v2.0.4 bump this PR was after, so both DoS fixes are already in. Closing — the diff here would be empty against current main.

PR #1084, opened 2026-08-05
Support URL quick replies on Telegram as inline keyboard link buttons

norkans7:
URL quick replies were dropped for Telegram because reply keyboard buttons can't open links. Inline keyboard URL buttons can, and render attached to the message on all Telegram clients, so they're used instead:

* when a message's quick replies are all of type `url`, they're sent as an `InlineKeyboardMarkup` of link buttons
* Telegram only allows one keyboard type per message, so URL replies mixed with types that need the reply keyboard (text, location, form) continue to be dropped with a logged channel error
* URL replies missing their URL are dropped with a "missing its extra value" error

No new webhook handling is needed — opening a link is entirely client-side.

### DIFF
--- a/handlers/telegram/handler.go
+++ b/handlers/telegram/handler.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


norkans7:
Replaced by #1087

PR #1085, opened 2026-08-05
Don't make Telegram keyboards single use when they contain form or location buttons

norkans7:
Form and location quick reply buttons launch a separate interaction — opening a Mini App or a location picker — that the user might not complete, and Telegram syncs one-time keyboard state across a user's devices, so a single tap on one client hides the keyboard everywhere. That made form keyboards effectively disappear before the user could open the form on another client.

Since keyboards are already explicitly cleared with `remove_keyboard` on the next message, keyboards containing form or location buttons now drop `one_time_keyboard` so they stay available until the conversation moves on.

Text-only keyboards keep the existing single-use behavior, where the tap itself is the reply.

### DIFF
--- a/handlers/telegram/handler_test.go
+++ b/handlers/telegram/handler_test.go
@@ -919,7 +919,7 @@ var outgoingCases = []OutgoingTestCase{
 			},
 		},
 		ExpectedRequests: []ExpectedRequest{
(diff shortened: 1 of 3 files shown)


norkans7:
Replaced by https://github.com/nyaruka/courier/pull/1087

PR #1089, opened 2026-08-05
Drop Telegram url quick replies with invalid URLs instead of failing the send

norkans7:
Telegram rejects an entire message with `400 BUTTON_URL_INVALID` if an inline keyboard button URL isn't valid, so a single malformed url quick reply would fail the whole send. Validate the URL when filtering and drop invalid ones with a logged channel error instead, so the message still goes out.

The validation approximates Telegram's rules for inline button URLs:

* `http://`, `https://` and `tg://` URLs are accepted (`tg://` deep links are valid button URLs per the Bot API)
* URLs containing whitespace are rejected
* HTTP(S) hostnames must contain a dot or be an IP address — Telegram's parser rejects bare hostnames like `localhost`

It's an approximation of an undocumented parser, so a genuinely exotic URL can still fail the send — this catches what can be caught cheaply.

Stacked on #1087.

### DIFF
--- a/handlers/telegram/handler.go
+++ b/handlers/telegram/handler.go
@@ -5,12 +5,14 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"slices"
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/nyaruka/courier/v26"
 	"github.com/nyaruka/courier/v26/core/models"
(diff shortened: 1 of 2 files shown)


norkans7:
A note on how the validation rules were chosen, since Telegram doesn't document its button URL parser:

Per the [Bot API](https://core.telegram.org/bots/api#inlinekeyboardbutton), the `url` field of an inline button accepts "HTTP or tg:// URL" — so `http://`, `https://` and `tg://` are the valid schemes. Known `BUTTON_URL_INVALID` triggers, and what this PR does about each:

| Case | Telegram | This PR |
|---|---|---|
| No scheme (`example.com`) | rejects | dropped with logged error |
| Unsupported scheme (`mailto:`, `ftp:`, …) | rejects | dropped with logged error |
| Empty URL | rejects | dropped (missing-extra error) |
| Bare hostname without TLD (`http://localhost`) | rejects — its parser requires a TLD, though IP addresses work ([tdesktop#7827](https://github.com/telegramdesktop/tdesktop/issues/7827)) | dropped with logged error |
| Unencoded whitespace | rejects | dropped with logged error |
| `tg://` deep links | **accepts** | passed through |
| IDN hosts, ports, IPv6 literals | accepts | passed through |

This is an approximation of an undocumented parser, so an exotic URL can still fail the send with `BUTTON_URL_INVALID` — the aim is to cheaply catch the mistakes flow authors will actually make while never dropping a URL Telegram would have accepted.

PR #1090, opened 2026-08-05
Reuse FilterSupportedQuickReplies for the WhatsApp and Turn quick reply buckets

norkans7:
Replaces the two WhatsApp-specific filter functions (`FilterFormQuickReplies` and `FilterURLQuickReplies`, also used by the Turn handler) with a single `handlers.FilterSupportedQuickReplies` pass up front — which already does both the type filtering and the missing-extra drop with the same error wording — followed by cheap `FilterQuickRepliesByType` bucket splits.

No new helper needed, and behavior is unchanged for the four existing quick reply types. If a new type is ever added, these handlers will now log "isn't supported" like other handlers do instead of silently ignoring it.

### DIFF
--- a/handlers/meta/whataspp_test.go
+++ b/handlers/meta/whataspp_test.go
@@ -929,7 +929,7 @@ var whatsappOutgoingTests = []OutgoingTestCase{
 			Body: `{"messaging_product":"whatsapp","recipient_type":"individual","to":"250788123123","type":"text","text":{"body":"Interactive form msg","preview_url":false}}`,
 		}},
 		ExpectedExtIDs:    []string{"157b5e14568e8"},
-		ExpectedLogErrors: []*clogs.Error{{Message: "form quick reply is missing a form ID and can't be sent"}},
+		ExpectedLogErrors: []*clogs.Error{{Message: "quick reply of type form is missing its extra value and can't be sent"}},
 	},
 	{
 		Label:           "Interactive with URL button",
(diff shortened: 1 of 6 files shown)


rowanseymour (commented):
Like the direction, but I think we can do this without adding a new function — `FilterSupportedQuickReplies` already does both checks (type filtering *and* the missing-extra drop) with the exact same error message. It just needs to be called once up front instead of per-bucket:

```go
sqrs := handlers.FilterSupportedQuickReplies(msg.QuickReplies(), clog,
    models.QuickReplyTypeText, models.QuickReplyTypeLocation, models.QuickReplyTypeForm, models.QuickReplyTypeURL)
qrs := handlers.FilterQuickRepliesByType(sqrs, models.QuickReplyTypeText)
locationQRs := handlers.FilterQuickRepliesByType(sqrs, models.QuickReplyTypeLocation)
formQRs := handlers.FilterQuickRepliesByType(sqrs, models.QuickReplyTypeForm)
urlQRs := handlers.FilterQuickRepliesByType(sqrs, models.QuickReplyTypeURL)
```

That would let this PR delete `FilterQuickRepliesWithExtra` entirely and reuse the two existing helpers. The error string is already identical so the test updates here stay exactly as they are. Since those four types are all the types that exist, behavior is unchanged today — and if a fifth type is ever added, these handlers would log "isn't supported" like every other handler does instead of silently dropping it.

FWIW on the new function as written: for a type where `RequiresExtra()` is false it degenerates into plain `FilterQuickRepliesByType`, so the `qr.RequiresExtra() &&` guard is dead weight given it's only ever called with `form`/`url` — another sign the layering wants to be one supported-filter pass plus cheap bucket splits.

norkans7:
Done — reworked as suggested: one `FilterSupportedQuickReplies` pass with all four types up front, then `FilterQuickRepliesByType` bucket splits, in both the WhatsApp and Turn handlers. `FilterQuickRepliesWithExtra` is gone entirely, and the test updates stayed exactly as they were since the error wording is identical. Agreed on the dead `RequiresExtra()` guard — that layering smell is gone with the function.

PR #1107, opened 2026-08-11
Send Turn attachments by link instead of uploading them first

rowanseymour:
## Summary

Follow-up to #1105, which made Turn template header media go out as a `link`. That left the handler with two strategies for the same job: template headers referenced media by URL, while ordinary attachments were downloaded by courier and re-uploaded to `/v1/media` so the send could reference the returned media id.

This makes all outgoing attachments use `link`, matching what `handlers/meta/whatsapp` already does for the Cloud API, and deletes the upload machinery.

## Changes

- Attachments are sent as `{"link": ...}`; `buildPayloads` builds `&mediaObject{Link: mediaURL}` directly.
- `fetchMediaID` is gone, along with the 24h valkey media-id cache and the 15 minute in-process failure cache it consulted, and the `init()` that started the latter. The send path no longer touches valkey.
- `mediaObject.ID` dropped as dead. Inbound media resolution (`resolveMediaURL`, which turns an incoming Turn media id into a download URL) is untouched.
- `buildPayloads` loses its `ctx` and `*handler` parameters, which existed only to reach `fetchMediaID`.
- Imports no longer needed: `jsonparser`, `redis`, `cache`, `httpx`, `vkutil`.
- Removed a dead commented-out `whatsapp.GetMsgPayloads` line at the call site.

Net: +33 / −269 across two files.

## Behavior

Sending a single image attachment, before — three requests, the first two only to obtain an id:

```
GET  https://foo.bar/image.jpg
POST /v1/media
POST /v1/messages  {"to":"...","type":"image","image":{"id":"36c484d1-...","caption":"image caption"}}
```

After — one request:

```
POST /v1/messages  {"to":"...","type":"image","image":{"link":"https://foo.bar/image.jpg","caption":"image caption"}}
```

## Why this is safe

Sending by link was already the fallback on every failure path. `fetchMediaID` returned an empty id when the download or the upload failed, and the caller then left the link populated:

```go
if err == nil && mediaID != "" { mediaURL = "" }
mediaPayload := &mediaObject{ID: mediaID, Link: mediaURL}
```

The pre-existing tests make this explicit — the attachment cases in `defaultSendTestCases` mocked `/v1/media` returning 400 or an empty body and asserted the link body. They were already testing the path this PR makes the only one.

## Reviewer note

This requires that the configured API endpoint can fetch the attachment URLs itself. #1105 already relies on that for template headers; this extends it to all attachments, and there is no longer an upload path to fall back on if an operator's media storage isn't reachable from their endpoint. That is the one behavior change worth communicating to operators.

## Test plan

- Removed `mediaCacheSendTestCases` (6 cases covering upload success, upload failure, and media-id cache reuse) — all of it exercised machinery that no longer exists.
- Simplified the five attachment cases in `defaultSendTestCases` to drop the download and upload requests; they continue to assert the exact `link` request body for audio, document, image and video.
- Added a template case whose body component has no variables of its own, pinning the `"parameters":null` serialization introduced by #1105 rather than leaving it assumed.
- `go test -p=1 ./...` passes; `gofmt` clean.


norkans7:
I think we need to first upload the media and send the media ID 
https://whatsapp.turn.io/docs/api/messages#media-messages

rowanseymour:
You're right, and thanks for catching it — I've closed this.

Checking the section you linked: `link` is documented for media templated messages and for interactive message headers, but the media messages section shows only the `id` form and says the uploaded media ID must be provided. #1105 sits inside the documented boundary; this PR crossed it.

My argument for why the removal was safe doesn't hold up. I reasoned that sending by link was already the fallback whenever the upload failed, so making it the only path changed nothing. That proves the code path existed, not that the API accepts it. I also leaned on the existing attachment tests, which mock `/v1/media` failing and assert the link request body — but those mocks return a canned 201, so they only pin what we send, never what Turn accepts.

Worth a separate look: if link really is rejected for plain media messages, the current fallback has been quietly producing failed sends whenever an upload fails. `fetchMediaID` swallows the failure and returns an empty id, and the caller then ships a link payload. An upload failure should probably surface as a retryable error instead of degrading into a send that can't succeed.

Two things here don't depend on the wrong premise and I'll pull them into a small PR: the template test case that pins the `"parameters":null` serialization for a component with no variables, and the removal of a dead commented-out line at the send call site.

